PR #8, opened 2020-01-22
Fix metric parsing

SchSeba:
This PR fix an issue with metric parsing

Issue:
ptp4l[76560.366]: rms   25 max   28 freq +16596 +/-   1
E0122 09:28:28.924420  231135 metrics.go:110] ptp4l failed to extract metrics unknown output format

Fix:
We check now the len to find if we have a deply from master or not.

phc2sys[80512.201]: enp6s0 rms   20 max   20 freq +16540 +/-   0 delay  1183 +/-   0
E0122 10:40:07.737644  160830 metrics.go:110] phc2sys failed to extract metrics unknown output format

Fix:
We now remove eveything before the "rms" string

### DIFF
--- a/pkg/daemon/metrics.go
+++ b/pkg/daemon/metrics.go
@@ -103,32 +103,36 @@ func extractMetrics(processName, output string){
 }
 
 func extractSummaryMetrics(processName, output string) (offsetFromMaster, maxOffsetFromMaster, frequencyAdjustment, delayFromMaster float64) {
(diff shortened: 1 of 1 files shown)


SchSeba:
/cc @zshi-redhat 

can you please take a look?

zshi-redhat:
> /cc @zshi-redhat
> 
> can you please take a look?

@SchSeba  Thanks for capturing this issue and proposing the fix!
/lgtm

PR #15, opened 2020-07-14
Allow to use the ptp4lConf variable

SchSeba:
This PR allow to apply a custom ptp4l.conf file

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -3,6 +3,8 @@ package daemon
 import (
 	"bufio"
 	"fmt"
+	"io/ioutil"
+	"os"
 	"os/exec"
 	"strings"
 	"syscall"
@@ -16,6 +18,7 @@ import (
 const (
 	PtpNamespace         = "openshift-ptp"
 	PTP4L_CONF_FILE_PATH = "/etc/ptp4l.conf"
+	PTP4L_CONF_DIR       = "/ptp4l-conf"
 )
 
 // ProcessManager manages a set of ptpProcess
@@ -96,7 +99,6 @@ func (dn *Daemon) Run() {
 			return
 		}
 	}
-	return
 }
 
 func printWhenNotNil(p *string, description string) {
@@ -149,6 +151,20 @@ func applyNodePTPProfile(pm *ProcessManager, nodeProfile *ptpv1.PtpProfile) erro
 	}
 
 	if nodeProfile.Ptp4lOpts != nil && nodeProfile.Interface != nil {
+		if nodeProfile.Ptp4lConf != nil && *nodeProfile.Ptp4lConf != "" {
+			if _, err := os.Stat(PTP4L_CONF_DIR); os.IsNotExist(err) {
+				err = os.Mkdir(PTP4L_CONF_DIR, 0644)
+				if err != nil {
+					return fmt.Errorf("failed to create the ptp4l custom configuration directory: %v", err)
+				}
+			}
+
+			err := ioutil.WriteFile(fmt.Sprintf("%s/%s.conf", PTP4L_CONF_DIR, *nodeProfile.Name), []byte(*nodeProfile.Ptp4lConf), 0644)
+			if err != nil {
+				return fmt.Errorf("failed to write the configuration file named %s: %v", fmt.Sprintf("%s/%s.conf", PTP4L_CONF_DIR, *nodeProfile.Name), err)
+			}
+		}
+
 		pm.process = append(pm.process, &ptpProcess{
 			name:   "ptp4l",
 			exitCh: make(chan bool),
@@ -175,8 +191,13 @@ func phc2sysCreateCmd(nodeProfile *ptpv1.PtpProfile) *exec.Cmd {
 
 // ptp4lCreateCmd generate ptp4l command
 func ptp4lCreateCmd(nodeProfile *ptpv1.PtpProfile) *exec.Cmd {
+	confFilePath := PTP4L_CONF_FILE_PATH
+	if nodeProfile.Ptp4lConf != nil && *nodeProfile.Ptp4lConf != "" {
+		confFilePath = fmt.Sprintf("%s/%s.conf", PTP4L_CONF_DIR, *nodeProfile.Name)
+	}
+
 	cmdLine := fmt.Sprintf("/usr/sbin/ptp4l -f %s -i %s %s",
-		PTP4L_CONF_FILE_PATH,
+		confFilePath,
 		*nodeProfile.Interface,
 		*nodeProfile.Ptp4lOpts)
 
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -123,7 +123,7 @@ func main() {
 			ptpConfUpdate.NodeProfile = ptpConfig
 			ptpConfUpdate.UpdateCh <- true
 		case sig := <-sigCh:
-			glog.Infof("signal received, shutting down", sig)
+			glog.Info("signal received, shutting down", sig)
 			return
 		}
 	}


pkg/daemon/metrics.go
```diff
@@ -159,8 +159,8 @@ func addFlagsForMonitor(nodeProfile *ptpv1.PtpProfile) {
 		}
 
 		if !strings.Contains(*nodeProfile.Ptp4lOpts, "--summary_interval") {
-			glog.Info("adding --summary_interval 1 to print summary messages to stdout for ptp4l to use prometheus exporter")
-			*nodeProfile.Ptp4lOpts = fmt.Sprintf("%s --summary_interval 1", *nodeProfile.Ptp4lOpts)
+			glog.Info("adding --summary_interval 0 to print summary messages to stdout for ptp4l to use prometheus exporter")
```
zshi-redhat, line 162: Changing summary_interval to 0 will result a different format of ptp4l output message. will it affect how message is parsed?

pkg/daemon/daemon.go
```diff
@@ -149,6 +151,20 @@ func applyNodePTPProfile(pm *ProcessManager, nodeProfile *ptpv1.PtpProfile) erro
...
 	if nodeProfile.Ptp4lOpts != nil && nodeProfile.Interface != nil {
+		if nodeProfile.Ptp4lConf != nil && *nodeProfile.Ptp4lConf != "" {
+			if _, err := os.Stat(PTP4L_CONF_DIR); os.IsNotExist(err) {
+				err = os.Mkdir(PTP4L_CONF_DIR, 0644)
+				if err != nil {
+					return fmt.Errorf("failed to create the ptp4l custom configuration directory: %v", err)
+				}
+			}
+
+			err := ioutil.WriteFile(fmt.Sprintf("%s/%s.conf", PTP4L_CONF_DIR, *nodeProfile.Name), []byte(*nodeProfile.Ptp4lConf), 0644)
```
zshi-redhat, line 162: By writing the content of Ptp4lConf in a file, we are expecting user may assign the full content of ptp4l.conf to Ptp4lConf, instead of using it as a filesystem path?

SchSeba: right he needs to right the full file in the config.

maybe a later feature we can implement is using the default ptp config that exist inside the pod `/etc/ptp4l.conf` and merge them.

right now this is the example I used to test it
```yaml
apiVersion: ptp.openshift.io/v1
kind: PtpConfig
metadata:
  name: slave
  namespace: openshift-ptp
spec:
  profile:
  - name: "slave"
    interface: "ens1f0"
    ptp4lOpts: "-2 -s --summary_interval 0"
    phc2sysOpts: "-a -r"
    ptp4lConf: |
      [global]
      #
      # Default Data Set
      #
      twoStepFlag		1
      slaveOnly		0
      priority1		128
      priority2		128
      domainNumber		24
      #utc_offset		37
      clockClass		248
      clockAccuracy		0xFE
      offsetScaledLogVariance	0xFFFF
      free_running		0
      freq_est_interval	1
      dscp_event		0
      dscp_general		0
      dataset_comparison	ieee1588
      G.8275.defaultDS.localPriority	128
      #
      # Port Data Set
      #
      logAnnounceInterval	-3
      logSyncInterval		-4
      logMinDelayReqInterval	-4
      logMinPdelayReqInterval	-4
      announceReceiptTimeout	3
      syncReceiptTimeout	0
      delayAsymmetry		0
      fault_reset_interval	4
      neighborPropDelayThresh	20000000
      masterOnly		0
      G.8275.portDS.localPriority	128
      #
      # Run time options
      #
      assume_two_step		0
      logging_level		6
      path_trace_enabled	0
      follow_up_info		0
      hybrid_e2e		0
      inhibit_multicast_service	0
      net_sync_monitor	0
      tc_spanning_tree	0
      tx_timestamp_timeout	1
      unicast_listen		0
      unicast_master_table	0
      unicast_req_duration	3600
      use_syslog		1
      verbose			0
      summary_interval	0
      kernel_leap		1
      check_fup_sync		0
      #
      # Servo Options
      #
      pi_proportional_const	0.0
      pi_integral_const	0.0
      pi_proportional_scale	0.0
      pi_proportional_exponent	-0.3
      pi_proportional_norm_max	0.7
      pi_integral_scale	0.0
      pi_integral_exponent	0.4
      pi_integral_norm_max	0.3
      step_threshold		0.0
      first_step_threshold	0.00002
      max_frequency		900000000
      clock_servo		pi
      sanity_freq_limit	200000000
      ntpshm_segment		0
      #
      # Transport options
      #
      transportSpecific	0x0
      ptp_dst_mac		01:1B:19:00:00:00
      p2p_dst_mac		01:80:C2:00:00:0E
      udp_ttl			1
      udp6_scope		0x0E
      uds_address		/var/run/ptp4l
      #
      # Default interface options
      #
      clock_type		OC
      network_transport	UDPv4
      delay_mechanism		E2E
      time_stamping		hardware
      tsproc_mode		filter
      delay_filter		moving_median
      delay_filter_length	10
      egressLatency		0
      ingressLatency		0
      boundary_clock_jbod	0
      #
      # Clock description
      #
      productDescription	;;
      revisionData		;;
      manufacturerIdentity	00:00:00
      userDescription		;
      timeSource		0xA0
  recommend:
  - profile: "slave"
    priority: 4
    match:
    - nodeLabel: "ptp/slave"

```

pkg/daemon/metrics.go
SchSeba: you are right I will revert this and work on another PR to support both the summary and the regular output.

In Telco they will like to use the normal output without the summary.

pkg/daemon/daemon.go
fedepaol: Another thing to notice here is that that field (`Ptp4lConf`) wasn't used at all before.

zshi-redhat: Initially I was intended to support such a case that user may want to define their own ptp4l.conf (e.g. they can provide a new ptp4l.conf file via MCO and use this PtpProfile.Ptp4lConf option to specify where that file is located). And I'm not sure whether it is required to change the ptp4l.conf file because most of those commonly used options can be configured by setting `PtpProfile.ptp4lOpts` which overwrite the config in `ptp4l.conf`. 

It looks to me that maybe there are many options in ptp4l.conf user wants to change, so instead of specifying all of them in `PtpProfile.ptp4lOpts` , they would like to provide a full ptp4l.conf file. Am I understand the use case correct? @SchSeba .

SchSeba: > Initially I was intended to support such a case that user may want to define their own ptp4l.conf (e.g. they can provide a new ptp4l.conf file via MCO and use this PtpProfile.Ptp4lConf option to specify where that file is located). And I'm not sure whether it is required to change the ptp4l.conf file because most of those commonly used options can be configured by setting PtpProfile.ptp4lOpts which overwrite the config in ptp4l.conf.

I looked into the ptp operator and there is no mount point for the linuxptp-daemon pod with the host so I am not sure how you can request the MCO to create a file and the pod will be able to use it.

> It looks to me that maybe there are many options in ptp4l.conf user wants to change, so instead of specifying all of them in PtpProfile.ptp4lOpts , they would like to provide a full ptp4l.conf file. Am I understand the use case correct?

Right but not only to make it easy for the user there are many configurations that you can only change in the config file and are not exposed to the command line arguments.

zshi-redhat: > > Initially I was intended to support such a case that user may want to define their own ptp4l.conf (e.g. they can provide a new ptp4l.conf file via MCO and use this PtpProfile.Ptp4lConf option to specify where that file is located). And I'm not sure whether it is required to change the ptp4l.conf file because most of those commonly used options can be configured by setting PtpProfile.ptp4lOpts which overwrite the config in ptp4l.conf.
> 
> I looked into the ptp operator and there is no mount point for the linuxptp-daemon pod with the host so I am not sure how you can request the MCO to create a file and the pod will be able to use it.
> 

It was not added in daemon manifest as there was no such request to replace default /etc/ptp4l.conf
Another usage for this option is when adding timemaster support, it will contain the additional options copied to the generated ptp4l.conf file, but I'm not sure if this CRD will be re-used when supporting timemaster.

> > It looks to me that maybe there are many options in ptp4l.conf user wants to change, so instead of specifying all of them in PtpProfile.ptp4lOpts , they would like to provide a full ptp4l.conf file. Am I understand the use case correct?
> 
> Right but not only to make it easy for the user there are many configurations that you can only change in the config file and are not exposed to the command line arguments.

ack.

SchSeba:
@zshi-redhat there is something missing in my PR?

zshi-redhat:
> @zshi-redhat there is something missing in my PR?

nope, I just wait in case there are other comments, if not, I will lgtm at the end of today.

fedepaol:
> > @zshi-redhat there is something missing in my PR?
> 
> nope, I just wait in case there are other comments, if not, I will lgtm at the end of today.

I was waiting for yours, classical deadlock :-) Looks good to me.

zshi-redhat:
> > > @zshi-redhat there is something missing in my PR?
> > 
> > 
> > nope, I just wait in case there are other comments, if not, I will lgtm at the end of today.
> 
> I was waiting for yours, classical deadlock :-) Looks good to me.

unlock :-)
/lgtm

PR #20, opened 2020-09-10
Support multiple interfaces

SchSeba:
This PR needs https://github.com/openshift/ptp-operator/pull/67 in the operator to support multiple profiles

Base on the proposal https://github.com/openshift/enhancements/pull/431

### DIFF
--- a/pkg/daemon/config.go
+++ b/pkg/daemon/config.go
@@ -1,0 +1,67 @@
+package daemon
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/golang/glog"
+
+	ptpv1 "github.com/openshift/ptp-operator/pkg/apis/ptp/v1"
+)
+
+// LinuxPTPUpdate controls whether to update linuxPTP conf
+// and contains linuxPTP conf to be updated. It's rendered
+// and passed to linuxptp instance by daemon.
+type LinuxPTPConfUpdate struct {
+	UpdateCh               chan bool
+	NodeProfiles           []ptpv1.PtpProfile
+	appliedNodeProfileJson []byte
+}
+
+func (l *LinuxPTPConfUpdate) UpdateConfig(nodeProfilesJson []byte) error {
+	if string(l.appliedNodeProfileJson) == string(nodeProfilesJson) {
+		return nil
+	}
+
+	if nodeProfiles, ok := tryToLoadOldConfig(nodeProfilesJson); ok {
+		glog.Info("load profiles using old method")
+		l.appliedNodeProfileJson = nodeProfilesJson
+		l.NodeProfiles = nodeProfiles
+		l.UpdateCh <- true
+
+		return nil
+	}
+
+	if nodeProfiles, ok := tryToLoadConfig(nodeProfilesJson); ok {
+		glog.Info("load profiles")
+		l.appliedNodeProfileJson = nodeProfilesJson
+		l.NodeProfiles = nodeProfiles
+		l.UpdateCh <- true
+
+		return nil
+	}
+
+	return fmt.Errorf("unable to load profile config")
+}
+
+// Try to load the multiple policy config
+func tryToLoadConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
+	ptpConfig := []ptpv1.PtpProfile{}
+	err := json.Unmarshal(nodeProfilesJson, &ptpConfig)
+	if err != nil {
+		return nil, false
+	}
+
+	return ptpConfig, true
+}
+
+// For backward compatibility we also try to load the one policy scenario
+func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
+	ptpConfig := &ptpv1.PtpProfile{}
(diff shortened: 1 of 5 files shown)


pkg/daemon/daemon.go
```diff
@@ -131,57 +119,107 @@ func applyNodePTPProfile(pm *ProcessManager, nodeProfile *ptpv1.PtpProfile) erro
...
+
+	socketPath := fmt.Sprintf("/var/run/ptp4l.%d.socket", runID)
+
+	config := fmt.Sprintf("%s\nuds_address %s\nmessage_tag [%s]",
+		*nodeProfile.Ptp4lConf,
+		socketPath,
+		*nodeProfile.Interface)
+	nodeProfile.Ptp4lConf = &config
+
+	commandLine := fmt.Sprintf("%s -z %s -t [%s]",
```
zshi-redhat, line 214: There is a warning in phc2sys man page:

```       Be  cautious when the same configuration file is used for both ptp4l and phc2sys.  Keep in mind, that values specified in the
       configuration file take precedence over their default values. If a certain option, which is common to ptp4l and  phc2sys,  is
       specified  to  a  non-default value in the configuration file (p.e., for ptp4l), then this non-default value applies also for
       phc2sys. This might be not what is expected.

       It is recommended to use seperate configuration files for ptp4l and phc2sys in order to avoid any unexpected behavior.
```

Since we are defining different config files for each ptp4l instance, shall we also consider using the same ptp4l config file for phc2sys instance that is created along with the ptp4l instance?

zshi-redhat: If we use the same ptp4l.conf file for phc2sys and ptp4l instances created in one nodeProfile, then `-z` and `-t` options might not be needed. so this function can be used only for updating the ptp4l.conf.

pkg/daemon/daemon.go
```diff
@@ -131,57 +119,107 @@ func applyNodePTPProfile(pm *ProcessManager, nodeProfile *ptpv1.PtpProfile) erro
...
 
+func (dn *Daemon) addProfileConfig(runID int, nodeProfile *ptpv1.PtpProfile) string {
+	// TODO: later implement a merge capability
+	if nodeProfile.Ptp4lConf == nil || *nodeProfile.Ptp4lConf == "" {
+		// We need to copy this to another variable because is a pointer
+		config := string(dn.ptpUpdate.defaultPTP4lConfig)
+		nodeProfile.Ptp4lConf = &config
+	}
+
+	socketPath := fmt.Sprintf("/var/run/ptp4l.%d.socket", runID)
```
zshi-redhat, line 206: Maybe define both socketPath outside of addProfileConfig and pass it in as a param?
It doesn't sound right that addProfileConfig returns a socket path which is only used by phc2sys instance.

pkg/daemon/config.go
```diff
@@ -0,0 +1,87 @@
...
+
+	return &LinuxPTPConfUpdate{UpdateCh: make(chan bool), defaultPTP4lConfig: defaultPTP4lConfig}, nil
+}
+
+func (l *LinuxPTPConfUpdate) UpdateConfig(nodeProfilesJson []byte) error {
+	if string(l.appliedNodeProfileJson) == string(nodeProfilesJson) {
+		return nil
+	}
+
+	if nodeProfiles, ok := tryToLoadOldConfig(nodeProfilesJson); ok {
```
zshi-redhat, line 46: In which case would this be required? when operator and operand have mis-matched versions? like during upgrade period?

pkg/daemon/daemon.go
SchSeba: I took the implementation from the timemaster.

We are not pointing to the ptp4l config in the phc2sys program just to the socket.
Example:
```
/usr/sbin/ptp4l -m --summary_interval -4 -f nic1.conf -i eno1
/usr/sbin/phc2sys -a -r -m -u 0 -R 1.00 -z /var/run/ptp4l.0.socket  -t [0:eno1] -n 0

/usr/sbin/ptp4l -m --summary_interval -4 -f nic2.conf -i eno2
/usr/sbin/phc2sys -a -r -m -u 0 -R 1.00 -z /var/run/ptp4l.1.socket -t [0:eno2] -n 24
```

pkg/daemon/config.go
SchSeba: yes this can break if the operator is updated and also the operand but the reconcile on the policy takes more time to change to the new struct.

but let me change and first call the new method it will make the parse a bit faster as only at the beginning we can have the issue after that it should always use the new struct

pkg/daemon/daemon.go
zshi-redhat: Ack, so we are not using any ptp4l.conf file for phc2sys process.

Anther question: with multiple profiles be able to configured on one node, can we still always assume phc2sys and ptp4l come in pair? for example, will there be a profile that only contains phc2sys config `phc2sys -a -r -r`?

SchSeba: I think we can assume that but the program support the use-case that only one process starts if the options for one of the process is empty the linuxptp-daemon will not start that process (this was already supported I didn't change that)

SchSeba:
/hold 

until 4.7 branch is open

PR #23, opened 2020-10-28
Bug 1892467: Make the metrics parser more robust

SchSeba:
This commit add a check for the number of fields
in the metrics parser functions


zshi-redhat:
I think this patch fixes the issue in linked bug. but I'm not sure if there are other output messages which may break current conditional check on fields. we may need to revisit this if similar issue occurs again.

PR #24, opened 2020-10-29
Bug 1892724: Always replace the list of devices instead of appending.

fedepaol:
In CI there was a flake that added a VF to the list of devices in the NodePtpDevice CRD.
The daemon always append to the list of devices in the node status instead of replacing, so
if there is a race in the detection of a vf this is never reconciled.

### DIFF
--- a/pkg/daemon/ptpdev.go
+++ b/pkg/daemon/ptpdev.go
@@ -1,9 +1,10 @@
 package daemon
 
 import (
+	"time"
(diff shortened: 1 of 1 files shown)


fedepaol:
/hold
needs validation (at least of not breaking the current behaviour)

SchSeba (approved):
/lgtm

waiting for validation

PR #30, opened 2020-12-06
Add N3000 network nics discovery

SchSeba:
This commit add the avility to discovery the n3000 network nics as ptp capable nics

### DIFF
--- a/pkg/network/utils.go
+++ b/pkg/network/utils.go
@@ -83,14 +83,21 @@ func DiscoverPTPDevices() ([]string, error) {
 			continue
 		}
 
-		// sysfs address looks like: /sys/devices/pci0000:17/0000:17:02.0/0000:19:00.5/net/eno1
+		var PCIAddr string
 		pathSegments := strings.Split(link, "/")
-		if len(pathSegments) != 8 {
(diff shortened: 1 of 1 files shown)


pkg/network/utils.go
```diff
@@ -83,15 +83,17 @@ func DiscoverPTPDevices() ([]string, error) {
 			continue
 		}
 
-		// sysfs address looks like: /sys/devices/pci0000:17/0000:17:02.0/0000:19:00.5/net/eno1
+		var PCIAddr string
 		pathSegments := strings.Split(link, "/")
-		if len(pathSegments) != 8 {
+		if len(pathSegments) == 10 { // sysfs address for N3000 looks like: ../../devices/pci0000:85/0000:85:00.0/0000:86:00.0/0000:87:10.0/0000:8a:00.1/net/eth1
```
fedepaol, line 88: Would taking the len() - 3 make this more general purpouse / more robust?

Also, if we pick the wrong segment the following if will fail and we will add the device to the list.

Would it make sense to validate the existance of /sys/bus/pci/devices/PCIAddr too ?

PR #39, opened 2021-06-02
enable propagating metrics from ptp Linux daemon to sidecar via socket

aneeshkp:
This is for Openshift 4.9, moving metrics to the sidecar PTP plugin.
PTPLinux daemon will write cmd output to Unix socket which will be mounted by the deployment and shared between the sidecar and the ptp Linux daemon.
The PTP plugin will consume cmd out from the socket and create Prometheus metrics.
Do I need to keep metrics on Linux PTP Daemon set for backport purposes?

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"fmt"
 	"io/ioutil"
+	"net"
 	"os"
 	"os/exec"
 	"strings"
@@ -17,9 +18,11 @@ import (
 )
 
 const (
-	PtpNamespace         = "openshift-ptp"
-	PTP4L_CONF_FILE_PATH = "/etc/ptp4l.conf"
-	PTP4L_CONF_DIR       = "/ptp4l-conf"
+	PtpNamespace            = "openshift-ptp"
+	PTP4L_CONF_FILE_PATH    = "/etc/ptp4l.conf"
+	PTP4L_CONF_DIR          = "/ptp4l-conf"
+	connectionRetryInterval = 1 * time.Second
+	eventSocket             = "/cloud-native/events.sock"
 )
 
 // ProcessManager manages a set of ptpProcess
@@ -45,6 +48,8 @@ type Daemon struct {
 	// node name where daemon is running
 	nodeName  string
 	namespace string
+	// write logs to socket, this will also send metrics to the socket
+	stdoutToSocket bool
 
 	// kubeClient allows interaction with Kubernetes, including the node we are running on.
 	kubeClient *kubernetes.Clientset
@@ -62,15 +67,18 @@ type Daemon struct {
 func New(
 	nodeName string,
 	namespace string,
+	stdoutToSocket bool,
 	kubeClient *kubernetes.Clientset,
 	ptpUpdate *LinuxPTPConfUpdate,
 	stopCh <-chan struct{},
 ) *Daemon {
-	RegisterMetrics(nodeName)
-
+	if !stdoutToSocket {
+		RegisterMetrics(nodeName)
+	}
 	return &Daemon{
 		nodeName:       nodeName,
 		namespace:      namespace,
+		stdoutToSocket: stdoutToSocket,
 		kubeClient:     kubeClient,
 		ptpUpdate:      ptpUpdate,
 		processManager: &ProcessManager{},
@@ -143,7 +151,7 @@ func (dn *Daemon) applyNodePTPProfiles() error {
 	for _, p := range dn.processManager.process {
 		if p != nil {
 			time.Sleep(1 * time.Second)
-			go cmdRun(p)
+			go cmdRun(p, dn.stdoutToSocket)
 		}
 	}
 	return nil
@@ -151,7 +159,7 @@ func (dn *Daemon) applyNodePTPProfiles() error {
 
 func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile) error {
 	// This add the flags needed for monitor
-	addFlagsForMonitor(nodeProfile)
+	addFlagsForMonitor(nodeProfile, dn.stdoutToSocket)
 
 	socketPath := fmt.Sprintf("/var/run/ptp4l.%d.socket", runID)
 	// This will create the configuration needed to run the ptp4l and phc2sys
@@ -236,11 +244,17 @@ func ptp4lCreateCmd(nodeProfile *ptpv1.PtpProfile, confFilePath string) *exec.Cm
 }
 
 // cmdRun runs given ptpProcess and wait for errors
-func cmdRun(p *ptpProcess) {
+func cmdRun(p *ptpProcess, stdoutToSocket bool) {
 	glog.Infof("Starting %s...", p.name)
 	glog.Infof("%s cmd: %+v", p.name, p.cmd)
-
+	var c net.Conn
+	var err error
 	defer func() {
+		if stdoutToSocket && c != nil {
+			if err := c.Close(); err != nil {
+				glog.Errorf("closing connection returned error %s", err)
+			}
+		}
 		p.exitCh <- true
 	}()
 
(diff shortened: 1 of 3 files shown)


pkg/daemon/metrics.go
```diff
@@ -186,15 +186,18 @@ func extractRegularMetrics(processName, output string) (offsetFromMaster, maxOff
...
+func addFlagsForMonitor(nodeProfile *ptpv1.PtpProfile, stdoutToSocket bool) {
 	// If output doesn't exist we add it for the prometheus exporter
 	if nodeProfile.Phc2sysOpts != nil {
 		if !strings.Contains(*nodeProfile.Phc2sysOpts, "-m") {
 			glog.Info("adding -m to print messages to stdout for phc2sys to use prometheus exporter")
 			*nodeProfile.Phc2sysOpts = fmt.Sprintf("%s -m", *nodeProfile.Phc2sysOpts)
 		}
-
-		if !strings.Contains(*nodeProfile.Phc2sysOpts, "-u") {
+		// stdoutTOSocket is for side car to consume events, -u  will not generate logs with offset and clock state.
```
SchSeba, line 196: stdoutTOSocket -> stdoutToSocket

SchSeba: side car -> sidecar

pkg/daemon/metrics.go
```diff
@@ -186,15 +186,18 @@ func extractRegularMetrics(processName, output string) (offsetFromMaster, maxOff
...
+		// disable -u for  events
+		if stdoutToSocket && strings.Contains(*nodeProfile.Phc2sysOpts, "-u"){
+				glog.Error("-u option will not generate clock state events,  remove -u option")
```
SchSeba, line 199: will the sidecard work as expected?
if not we may need to abort starting the process it will be super hard to find this line in the linuxptp logs

pkg/daemon/metrics.go
```diff
@@ -186,15 +186,18 @@ func extractRegularMetrics(processName, output string) (offsetFromMaster, maxOff
...
+		}else if !stdoutToSocket && !strings.Contains(*nodeProfile.Phc2sysOpts, "-u") {
```
SchSeba, line 200: please run `go fmt`

pkg/daemon/daemon.go
```diff
@@ -247,6 +250,13 @@ func cmdRun(p *ptpProcess) {
 	//
 	// don't discard process stderr output
 	//
+	c, err := net.Dial("unix", "/tmp/metrics.sock")
+	if err != nil {
+		glog.Errorf("Dial error", err)
+		return
+	}
+	defer c.Close()
```
SchSeba, line 258: better to put this in the defer already exist above

pkg/daemon/daemon.go
```diff
@@ -247,6 +250,13 @@ func cmdRun(p *ptpProcess) {
 	//
 	// don't discard process stderr output
 	//
+	c, err := net.Dial("unix", "/tmp/metrics.sock")
```
SchSeba, line 253: can we have a race here if the sidecar doesn't start the service and we try to dial?

pkg/daemon/daemon.go
```diff
@@ -261,7 +271,15 @@ func cmdRun(p *ptpProcess) {
 		for scanner.Scan() {
 			output := scanner.Text()
 			fmt.Printf("%s\n", output)
-			extractMetrics(p.name, p.iface, output)
+			out := fmt.Sprintf("%s\n", output)
```
SchSeba, line 274: nit: configure out before the fmt.Printf and use it there

pkg/daemon/daemon.go
```diff
@@ -261,7 +271,15 @@ func cmdRun(p *ptpProcess) {
...
+			if stdoutToSocket {
+				_, err := c.Write([]byte(out))
```
SchSeba, line 276: if the sidecar pod crash/get restarted will the connection continue to work?

pkg/daemon/daemon.go
aneeshkp: possible but not likely, because ptp-config has to be configured before it could start to dial.

pkg/daemon/metrics.go
aneeshkp: sidecar won't have this log information and it will ignore generating events. but even better I can catch this information in sidecar by looking at the log for  `rms`  text and print an error or info.

pkg/daemon/daemon.go
aneeshkp: you are right, it won't connect again. end up as broken pipe.
 I don't have a reconnect logic. I should implement reconnect here.

pkg/daemon/daemon.go
aneeshkp: i am adding  logic to wait for the connection to be established first

pkg/daemon/metrics.go
aneeshkp: done

pkg/daemon/daemon.go

pkg/daemon/daemon.go

pkg/daemon/daemon.go

pkg/daemon/daemon.go

pkg/daemon/daemon.go
```diff
@@ -254,14 +263,41 @@ func cmdRun(p *ptpProcess) {
...
-			extractMetrics(p.name, p.iface, output)
+			if stdoutToSocket {
+				out := fmt.Sprintf("%s\n", output)
+				fmt.Printf("%s", out)
+				if connected {
+					_, err := c.Write([]byte(out))
+					if err != nil {
+						connected = false
+						glog.Errorf("Write error:", err)
+						go handleConnection()
```
SchSeba, line 294: Can you explain me here how this will work?
will you be able to continue using the c.Write after the second go routine will be connected?

pkg/daemon/daemon.go
```diff
@@ -254,14 +263,41 @@ func cmdRun(p *ptpProcess) {
...
 	done := make(chan struct{})
 
 	scanner := bufio.NewScanner(cmdReader)
 	go func() {
 		for scanner.Scan() {
 			output := scanner.Text()
-			fmt.Printf("%s\n", output)
-			extractMetrics(p.name, p.iface, output)
+			if stdoutToSocket {
+				out := fmt.Sprintf("%s\n", output)
```
SchSeba, line 287: you need to move this outside of the if because you remove the /n from the else print

SchSeba (commented):
two small comments.

I need to read this more closely we have a lot of go routines now.

General comment how the sidecar will work when multiple policies will be apply on one node?
Is the sidecard only report a nic status or its done base on the ptpConfig name?

pkg/daemon/daemon.go
```diff
@@ -62,15 +66,15 @@ type Daemon struct {
 func New(
 	nodeName string,
 	namespace string,
+	stdoutToSocket bool,
 	kubeClient *kubernetes.Clientset,
 	ptpUpdate *LinuxPTPConfUpdate,
 	stopCh <-chan struct{},
 ) *Daemon {
-	RegisterMetrics(nodeName)
```
SchSeba, line 69: I don't see any call for this function.
we will need this if the sidecar is not enabled no?

pkg/daemon/daemon.go
```diff
@@ -282,6 +318,15 @@ func cmdRun(p *ptpProcess) {
 	return
 }
 
+func handleConnection() (net.Conn, error) {
+	c, err := net.Dial("unix", "/cloud-native/events.sock")
```
SchSeba, line 322: nit: please put the link in a const

pkg/daemon/daemon.go
```diff
@@ -256,15 +270,51 @@ func cmdRun(p *ptpProcess) {
...
+		}()
+	} else {
+		go func(io.ReadCloser) {
+			for {
+			connect:
+				select {
+				case <-p.exitCh:
+					done <- struct{}{}
+				default:
+					c, err = net.Dial("unix", "/cloud-native/events.sock")
```
SchSeba, line 291: please reuse the function you already have `handleConnection`

pkg/daemon/daemon.go
```diff
@@ -282,6 +332,15 @@ func cmdRun(p *ptpProcess) {
 	return
 }
 
+func handleConnection() (c net.Conn, err error) {
+	c, err = net.Dial("unix", "/cloud-native/events.sock")
```
SchSeba, line 336: nit: add the socket as a conts

pkg/daemon/daemon.go
```diff
@@ -256,15 +270,51 @@ func cmdRun(p *ptpProcess) {
...
+					if err != nil {
+						glog.Errorf("error trying to connect to event socket. retrying in %d secs", connectionRetryInterval)
+						time.Sleep(connectionRetryInterval)
+						continue
+					}
+				}
+				scanner := bufio.NewScanner(cmdReader)
+				for scanner.Scan() {
+					output := scanner.Text()
+					if stdoutToSocket {
```
SchSeba, line 301: you already have `stdoutToSocket` you can remove the `else section

pkg/daemon/daemon.go
```diff
@@ -256,15 +270,51 @@ func cmdRun(p *ptpProcess) {
...
+							glog.Errorf("Write error:", err)
+							goto connect
+						}
+					} else {
+						fmt.Printf("%s\n", output)
+						extractMetrics(p.name, p.iface, output)
+					}
+				}
+			}
+			done <- struct{}{}
```
SchSeba, line 315: this sections is `Unreachable code`

pkg/daemon/daemon.go
```diff
@@ -256,15 +270,51 @@ func cmdRun(p *ptpProcess) {
...
+		}(cmdReader)
```
SchSeba, line 316: you don't need this you can just use it same as here

```
go func() {
			for scanner.Scan() {
				output := scanner.Text()
				fmt.Printf("%s\n", output)
				extractMetrics(p.name, p.iface, output)
			}
			done <- struct{}{}
		}()
```

pkg/daemon/daemon.go
aneeshkp: i delete that function

aneeshkp: function is not required ,since the code is  used only  here

PR #48, opened 2021-07-28
Bug 1985773: Update interface handling

josephdrichard:


pkg/daemon/daemon.go
```diff
@@ -42,6 +42,16 @@ type ptpProcess struct {
...
 
+type ptp4lconfSection struct {
+	name            string
+	options          []string
+}
+
+type ptp4lconf struct {
+	sections        map[string]ptp4lconfSection
+	ifaceString     string
+}
```
oribon, line 53: can you declare these and the relevant methods in config.go?

oribon (commented):
Hey, can you make the commit message more descriptive please?
btw, if you try using toml marshaling/unmarshaling this website can maybe be helpful https://xuri.me/toml-to-go/
Thanks

pkg/daemon/daemon.go
```diff
@@ -200,6 +210,41 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 	return nil
 }
 
+func populatePtp4lConf(tomlContent *string, output *ptp4lconf) {
```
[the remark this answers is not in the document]
josephdrichard: I will update the commit message.  I tried BurntSushi toml parser, but it doesn't work as the config file isn't valid toml (because of using spaces instead of =).  I would be more worried of unknowingly breaking something if I try switching " " to "=" to parse, and then switch back after.

pkg/daemon/daemon.go
josephdrichard: Will update in next revision.

pkg/daemon/daemon.go
```diff
@@ -200,6 +210,41 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
...
+			section.options = append(section.options, line)
+			output.sections[current] = section
+		}
+	}
+	output.ifaceString = strings.ReplaceAll(output.ifaceString, "][", ",")
+	output.ifaceString = strings.ReplaceAll(output.ifaceString, "[", "")
+	output.ifaceString = strings.ReplaceAll(output.ifaceString, "]", "")
+}
+
+func renderPtp4lConf(tomlOutput *string, input *ptp4lconf) {
```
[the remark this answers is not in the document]
josephdrichard: I am erasing the previous content to put the rendered content here.

pkg/daemon/daemon.go
```diff
@@ -200,6 +210,41 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
...
 
+func populatePtp4lConf(tomlContent *string, output *ptp4lconf) {
+	lines := strings.Split(*tomlContent, "\n")
+	var current string;
+	output.sections = map[string]ptp4lconfSection{}
+
+	for _, line := range lines{
+		if strings.HasPrefix(line, "["){
+			current = line
+			output.sections[current] = *new(ptp4lconfSection)
```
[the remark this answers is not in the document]
josephdrichard: I don't understand this comment.  I am allocating the new section, that will contain either the global or an interface config section.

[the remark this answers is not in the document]
oribon: it's more of a cosmetic change, we usually use &type{} instead of new() for new objects.
so in this case you should use and it won't change anything
```
output.sections[current] = &ptp4lconfSection{}
```

pkg/daemon/config.go
```diff
@@ -92,3 +103,38 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
 
 	return []ptpv1.PtpProfile{*ptpConfig}, true
 }
+
+func populatePtp4lConf(config *string, output *ptp4lconf) {
```
oribon, line 107: same as before, I think you don't need to use *string, string should be enough.
and about changing the functions to methods it could be a bit more clear than the current implementation,
i.e func (p *ptp4lconf) Populate(config string) { ...}

pkg/daemon/config.go
```diff
@@ -92,3 +103,38 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
...
+		}
+	}
+	output.ifaceString = strings.ReplaceAll(output.ifaceString, "][", ",")
+	output.ifaceString = strings.ReplaceAll(output.ifaceString, "[", "")
+	output.ifaceString = strings.ReplaceAll(output.ifaceString, "]", "")
+}
+
+func renderPtp4lConf(config *string, input *ptp4lconf) {
+
+	*config = ""
```
oribon, line 132: same as before, there is no use having config as input and immediately assigning it ""
you could declare this func as 
```
func (p *ptp4lconf) Render() string { ...}
```

josephdrichard: config is the output when rendering the config based off of the input ptp4lconf.

pkg/daemon/daemon.go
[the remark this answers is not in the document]
josephdrichard: updated

pkg/daemon/config.go
josephdrichard: updated

pkg/daemon/config.go
josephdrichard: updated

pkg/daemon/config.go
SchSeba: Hi I don't see the update here its still a function that is not part of a struct like this `func (p *ptp4lconf) Populate(config string) { ...}`

pkg/daemon/config.go
SchSeba: same maybe you didn't push? :)

pkg/daemon/daemon.go
```diff
@@ -42,6 +42,7 @@ type ptpProcess struct {
 	cmd             *exec.Cmd
 }
 
+
```
SchSeba, line 45: please remove this empty line

pkg/daemon/daemon.go
```diff
@@ -214,23 +215,40 @@ func (dn *Daemon) addProfileConfig(socketPath string, nodeProfile *ptpv1.PtpProf
...
 	}
 
-	config := fmt.Sprintf("%s\nuds_address %s\nmessage_tag [%s]",
-		*nodeProfile.Ptp4lConf,
-		socketPath,
-		*nodeProfile.Interface)
-	nodeProfile.Ptp4lConf = &config
+	output = populatePtp4lConf(nodeProfile.Ptp4lConf)
+
+	if nodeProfile.Interface != nil && *nodeProfile.Interface != "" {
```
SchSeba, line 228: can't you use the if in line 219?

pkg/daemon/daemon.go
```diff
@@ -214,23 +215,40 @@ func (dn *Daemon) addProfileConfig(socketPath string, nodeProfile *ptpv1.PtpProf
...
+		ifaceSection := fmt.Sprintf("[%s]\n", *nodeProfile.Interface)
+		output.sections[ifaceSection] = *new(ptp4lconfSection)
+	}
+
+
```
SchSeba, line 233: remove 1 empty line please

pkg/daemon/daemon.go
```diff
@@ -214,23 +215,40 @@ func (dn *Daemon) addProfileConfig(socketPath string, nodeProfile *ptpv1.PtpProf
...
+	uds_address := fmt.Sprintf("uds_address %s\n", socketPath)
+	message_tag := fmt.Sprintf("message_tag [%s]\n", output.ifaceString)
+	section := output.sections["[global]"]
+	section.options = append(section.options, message_tag)
+	section.options = append(section.options, uds_address)
+	output.sections["[global]"] = section
+
+
```
SchSeba, line 241: remove this 2 empty lines please no need for 3 empty lines

pkg/daemon/daemon.go
```diff
@@ -214,23 +215,40 @@ func (dn *Daemon) addProfileConfig(socketPath string, nodeProfile *ptpv1.PtpProf
...
+
+	*nodeProfile.Ptp4lConf = renderPtp4lConf(nodeProfile.Ptp4lConf, &output)
```
SchSeba, line 243: this needs to be `err := output.renderPtp4lConf(nodeProfile.Ptp4lConf)` then you can just use 

output.conf  and output.ifaceString

pkg/daemon/daemon.go
josephdrichard: Needs to be after populating the conf

SchSeba:
Hi please rebase

SchSeba:
/cc @aneeshkp 

please review this PR that is inline with the metrics feature we sync offline

pkg/daemon/config.go
```diff
@@ -21,6 +22,18 @@ type LinuxPTPConfUpdate struct {
 	defaultPTP4lConfig     []byte
 }
 
+type ptp4lconfSection struct {
+	name            string
+	options         []string
```
SchSeba, line 27: I think is better for this to be a `map[string]string`

pkg/daemon/daemon.go
```diff
@@ -214,23 +214,46 @@ func (dn *Daemon) addProfileConfig(socketPath string, nodeProfile *ptpv1.PtpProf
...
-		*nodeProfile.Interface)
-	nodeProfile.Ptp4lConf = &config
+	output.populatePtp4lConf(nodeProfile.Ptp4lConf)
+
+	if nodeProfile.Interface != nil && *nodeProfile.Interface != "" {
+		ifaceSection := fmt.Sprintf("[%s]\n", *nodeProfile.Interface)
+		output.sections[ifaceSection] = *new(ptp4lconfSection)
+	}
+
+	for n, line := range output.sections["[global]"].options{
```
SchSeba, line 232: better for the options to be a map it will make this easier

pkg/daemon/daemon.go
```diff
@@ -214,23 +214,46 @@ func (dn *Daemon) addProfileConfig(socketPath string, nodeProfile *ptpv1.PtpProf
...
+			output.sections["[global]"].options[n] = ""
+		}
+
+	}
+
+
+	uds_address := fmt.Sprintf("uds_address %s\n", socketPath)
+	message_tag := fmt.Sprintf("message_tag [%s]\n", *nodeProfile.Name)
+	section := output.sections["[global]"]
+	section.options = append(section.options, message_tag)
```
SchSeba, line 243: use a map as the options this will allow you to find the key (or create of not exist)

pkg/daemon/daemon.go
```diff
@@ -214,23 +214,46 @@ func (dn *Daemon) addProfileConfig(socketPath string, nodeProfile *ptpv1.PtpProf
...
+	*nodeProfile.Ptp4lConf = output.conf
+	*nodeProfile.Interface = output.ifaceString
 
 	commandLine := fmt.Sprintf("%s -z %s -t [%s]",
 		*nodeProfile.Phc2sysOpts,
 		socketPath,
-		*nodeProfile.Interface)
+		*nodeProfile.Name)
 	nodeProfile.Phc2sysOpts = &commandLine
+
```
SchSeba, line 256: no need for a new line here

pkg/daemon/metrics.go
```diff
@@ -95,13 +95,13 @@ func updatePTPMetrics(process, iface string, offsetFromMaster, maxOffsetFromMast
 }
 
 // extractMetrics ...
-func extractMetrics(processName, iface, output string) {
+func extractMetrics(processName string, iface []string, output string) {
 	if strings.Contains(output, " max ") {
 		offsetFromMaster, maxOffsetFromMaster, frequencyAdjustment, delayFromMaster := extractSummaryMetrics(processName, output)
-		updatePTPMetrics(processName, iface, offsetFromMaster, maxOffsetFromMaster, frequencyAdjustment, delayFromMaster)
+		updatePTPMetrics(processName, iface[0], offsetFromMaster, maxOffsetFromMaster, frequencyAdjustment, delayFromMaster)
```
SchSeba, line 101: why you put here only the first interface?
I think you need to change this to support multiple interfaces metrics

/cc @aneeshkp

josephdrichard: I didn't want to change the signature of that function here.  I expect Aneesh will update this when he changes updatePTPMetrics.

pkg/daemon/config.go
```diff
@@ -92,3 +105,42 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
...
+
+func (conf *ptp4lconf) populatePtp4lConf(config *string) {
+	lines := strings.Split(*config, "\n")
+	var current string;
+	conf.sections = make(map[string]ptp4lconfSection)
+
+	for _, line := range lines{
+		if strings.HasPrefix(line, "["){
+			current = line
+			current := strings.Split(line, "]")[0]
```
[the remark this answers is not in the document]
josephdrichard: I will add a check to make sure it's > 0.  `[global] #[global]` is valid so I don't want to fail if >1.

pkg/daemon/daemon.go
```diff
@@ -214,22 +215,34 @@ func (dn *Daemon) addProfileConfig(socketPath string, nodeProfile *ptpv1.PtpProf
...
+	} else {
+		output.ifaceString = fmt.Sprintf("[%s]", *nodeProfile.Interface)
 	}
 
-	config := fmt.Sprintf("%s\nuds_address %s\nmessage_tag [%s]",
-		*nodeProfile.Ptp4lConf,
-		socketPath,
-		*nodeProfile.Interface)
-	nodeProfile.Ptp4lConf = &config
+	output.populatePtp4lConf(nodeProfile.Ptp4lConf)
```
SchSeba, line 226: this function return an error but you are not doing anything with it

pkg/daemon/config.go
```diff
@@ -92,3 +106,55 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
 
 	return []ptpv1.PtpProfile{*ptpConfig}, true
 }
+
+func (conf *ptp4lConf) populatePtp4lConf(config *string) error {
```
SchSeba, line 110: If the user only configure interfaces in the conf (with no global section) we need to create a global I think from the file inside the container

pkg/daemon/daemon.go
```diff
@@ -200,7 +201,7 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 	return nil
 }
 
-func (dn *Daemon) addProfileConfig(socketPath string, nodeProfile *ptpv1.PtpProfile) {
+func (dn *Daemon) addProfileConfig(socketPath string, nodeProfile *ptpv1.PtpProfile, configFile string) {
```
SchSeba, line 204: nit: switch the string to be together 
```golang
socketPath, configFile  string, nodeProfile *ptpv1.PtpProfile
```

pkg/daemon/config.go
```diff
@@ -92,3 +106,55 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
...
+			return errors.New("Config option not in section")
+		}
+	}
+	conf.ifaceString = strings.Join(conf.mapping, ",")
+	return nil
+}
+
+func (conf *ptp4lConf) renderPtp4lConf() {
+
+	conf.conf = ""
```
SchSeba, line 152: no need for conf we are not using it at all.
this function just need to return a string and that's it

pkg/daemon/daemon.go
```diff
@@ -214,22 +215,34 @@ func (dn *Daemon) addProfileConfig(socketPath string, nodeProfile *ptpv1.PtpProf
...
+		ifaceSection := fmt.Sprintf("[%s]\n", *nodeProfile.Interface)
+		output.sections[ifaceSection] = *new(ptp4lConfSection)
+	}
+
+	section := output.sections["[global]"]
+	section.options["message_tag"] = fmt.Sprintf("[%s]", configFile)
+	section.options["uds_address"] = socketPath
+	output.sections["[global]"] = section
+
+	output.renderPtp4lConf()
```
SchSeba, line 238: lets make this function to return a string no need to save it in the conf variable

pkg/daemon/daemon.go
```diff
@@ -288,7 +296,7 @@ func cmdRun(p *ptpProcess, stdoutToSocket bool) {
 			for scanner.Scan() {
 				output := scanner.Text()
 				fmt.Printf("%s\n", output)
-				extractMetrics(p.name, p.iface, output)
+				extractMetrics(p.name, p.ifaces[0], output)
```
SchSeba, line 299: nit: please add a TODO to fix this in a following PR to fix the metrics

SchSeba (commented):
Thanks for the contribution some comments

pkg/daemon/config.go
aneeshkp: Just a thought , we could use RegEx to fetch all interfaces 
```
sectionHead := regexp.MustCompile(`\[([^\[\]]*)\]`)
	matches := sectionHead.FindAllStringSubmatch(*p.Ptp4lConf, -1)
	for _, v := range matches {
		if v[1] != ignorePtp4lSection && v[1] != singleInterface {
			interfaces = append(interfaces, &v[1])
		}
	}
	```

Comment (aneeshkp) [reply pkg/daemon/config.go:110]: ignorePtp4lSection="[gloabl]"

File: pkg/daemon/daemon.go (hunk 22 above)
Comment (josephdrichard) [reply pkg/daemon/daemon.go:238]: I need to update to do mapping when rendering in order to pick up ptpconfig interface in interface list, so need to use conf variable anyways.

File: pkg/daemon/config.go [hunk 24]
```diff
@@ -92,3 +106,59 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
...
+		if strings.HasPrefix(line, "[") {
+			currentSection = line
+			currentLine := strings.Split(line, "]")
+
+			if len(currentLine) < 2 {
+				return errors.New("Section missing closing ']'")
+			}
+
+			currentSection = fmt.Sprintf("%s]", currentLine[0])
+			section := ptp4lConfSection{}
```
SchSeba, line 125: add also the `name` field here

something like this no?

```golang
 section := ptp4lConfSection{name: currentSection, options: map[string]string{}}
```

pkg/daemon/config.go
```diff
@@ -92,3 +106,59 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
...
+			if split > 0 {
+				section := conf.sections[currentSection]
+				section.options[line[:split]] = line[split:]
+				conf.sections[currentSection] = section
+			}
+		} else {
+			return errors.New("Config option not in section")
+		}
+	}
+	if conf.sections["[global]"].options == nil {
```
SchSeba, line 139: _, exist := output.sections["[global]"]
if !exist {
output.sections["[global]"] = ptp4lConfSection{name: "[global]", options: map[string]string{}}
}

pkg/daemon/config.go
```diff
@@ -92,3 +106,59 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
...
+		section := ptp4lConfSection{}
+		section.options = make(map[string]string)
+		conf.sections["[global]"] = section
+	}
+	return nil
+}
+
+func (conf *ptp4lConf) renderPtp4lConf() {
```
SchSeba, line 147: please change this function to return the ifaceString and the config.

This way you can remove them from the struct there is no need to save the information

pkg/daemon/daemon.go
```diff
@@ -214,23 +218,36 @@ func (dn *Daemon) addProfileConfig(socketPath string, nodeProfile *ptpv1.PtpProf
...
-		*nodeProfile.Ptp4lConf,
-		socketPath,
-		*nodeProfile.Interface)
-	nodeProfile.Ptp4lConf = &config
+	section := output.sections["[global]"]
+	section.options["message_tag"] = fmt.Sprintf("[%s]", configFile)
+	section.options["uds_address"] = socketPath
+	output.sections["[global]"] = section
+
+	output.renderPtp4lConf()
```
SchSeba, line 240: this function can return 2 strings ptp4lConf and the interface list.

There is no need have them as fields in the struct of the ptp4lConf

SchSeba (commented):
Almost there!

just small changes and this PR should be ready

pkg/daemon/config.go
```diff
@@ -92,3 +106,59 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
...
+	lines := strings.Split(*config, "\n")
+	var currentSection string
+	conf.sections = make(map[string]ptp4lConfSection)
+
+	for _, line := range lines {
+		if strings.HasPrefix(line, "[") {
+			currentSection = line
+			currentLine := strings.Split(line, "]")
+
+			if len(currentLine) < 2 {
```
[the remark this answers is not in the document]
josephdrichard: Checking that it split into more than one string, meaning line has a closing ].

PR #53, opened 2021-08-31
Bug 1999185: with master summary_interval phc2sys adds sys to the log

aneeshkp:
log out put 
```
ptp4l.0.config master offset -2162130      s2    freq +22451884  delay 374976
ptp4l.0.config eno1   sys    offset        13     s2   freq      +6010 delay    468
```
Offset generally seen at index 2, if it is at index 3 then move it to index 2, but for some setup log adds sys field which throws an parsing error .
This PR makes sure the 2nd field is always offset. If it is found  at field 3 then  log an error

### DIFF
--- a/pkg/daemon/metrics.go
+++ b/pkg/daemon/metrics.go
@@ -26,6 +26,9 @@ const (
 	phc2sysProcessName = "phc2sys"
 	clockRealTime      = "CLOCK_REALTIME"
 	master             = "master"
+
+	offset = "offset"
+	rms    = "rms"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/daemon/metrics.go
```diff
@@ -309,6 +312,11 @@ func extractRegularMetrics(configName, processName, output string) (iface, clock
 		return
 	}
 
+	if fields[2] != offset {
+		glog.Errorf("%s failed to parse offset from master output %s error %s", processName, fields[1], "offset is not in right order")
```
SchSeba, line 316: maybe print the `output` here so we can have the full line?

SchSeba: no need because we print the line in the log

SchSeba (commented):
/lgtm
/approve

PR #60, opened 2021-11-19
Pull in latest ptp operator

josephdrichard:
Update vendored go module to latest.
This is necessary in order to pull in updated ptp profile type.
Also make required changes to match ptp operator where possible.


josephdrichard:
@SchSeba @aneeshkp I want to get this merged first before doing the change to support scheduling.
I had to update as some things were missing so couldn't just go vendor easily.  My intent was to match the ptp-operator wherever possible.

PR #61, opened 2021-11-20
Add support for changing ptp process scheduling

josephdrichard:
Add support for handling PtpSchedulingPolicy and PtpSchedulingPriority,
which are used to configure SCHED_FIFO scheduling.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -109,9 +109,18 @@ func (dn *Daemon) Run() {
 	}
 }
 
-func printWhenNotNil(p *string, description string) {
-	if p != nil {
-		glog.Info(description, ": ", *p)
+func printWhenNotNil(p interface{}, description string) {
+	switch v := p.(type) {
+	case *string:
+		if v != nil {
+			glog.Info(description, ": ", *v)
+		}
+	case *int64:
+		if v != nil {
+			glog.Info(description, ": ", *v)
+		}
+	default:
+		glog.Info(description, ": ", v)
 	}
 }
 
@@ -176,6 +185,8 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile) 
 	printWhenNotNil(nodeProfile.Ptp4lOpts, "Ptp4lOpts")
 	printWhenNotNil(nodeProfile.Ptp4lConf, "Ptp4lConf")
 	printWhenNotNil(nodeProfile.Phc2sysOpts, "Phc2sysOpts")
+	printWhenNotNil(nodeProfile.PtpSchedulingPolicy, "PtpSchedulingPolicy")
+	printWhenNotNil(nodeProfile.PtpSchedulingPriority, "PtpSchedulingPriority")
 	glog.Infof("------------------------------------")
 
 	if nodeProfile.Phc2sysOpts != nil {
@@ -253,9 +264,30 @@ func (dn *Daemon) addProfileConfig(socketPath string, configFile string, nodePro
 	return nil
 }
 
+// Add fifo scheduling if specified in nodeProfile
+func addScheduling(nodeProfile *ptpv1.PtpProfile, cmdLine string) string {
+	if nodeProfile.PtpSchedulingPolicy != nil && *nodeProfile.PtpSchedulingPolicy == "SCHED_FIFO" {
+		if nodeProfile.PtpSchedulingPriority == nil {
+			glog.Errorf("Priority must be set for SCHED_FIFO; using default scheduling.")
+			return cmdLine
+		}
+		priority := *nodeProfile.PtpSchedulingPriority
+		if priority < 1 || priority > 65 {
+			glog.Errorf("Invalid priority %d; using default scheduling.", priority)
+			return cmdLine
+		}
+		cmdLine = fmt.Sprintf("/bin/chrt -f %d %s", priority, cmdLine)
+		glog.Infof(cmdLine)
+		return cmdLine
+	}
+	return cmdLine
+}
+
 // phc2sysCreateCmd generate phc2sys command
 func phc2sysCreateCmd(nodeProfile *ptpv1.PtpProfile) *exec.Cmd {
 	cmdLine := fmt.Sprintf("/usr/sbin/phc2sys %s", *nodeProfile.Phc2sysOpts)
+	cmdLine = addScheduling(nodeProfile, cmdLine)
+
 	args := strings.Split(cmdLine, " ")
 	return exec.Command(args[0], args[1:]...)
 }
@@ -265,6 +297,7 @@ func ptp4lCreateCmd(nodeProfile *ptpv1.PtpProfile, confFilePath string) *exec.Cm
 	cmdLine := fmt.Sprintf("/usr/sbin/ptp4l -f %s %s",
 		confFilePath,
 		*nodeProfile.Ptp4lOpts)
+	cmdLine = addScheduling(nodeProfile, cmdLine)
 
 	args := strings.Split(cmdLine, " ")
 	return exec.Command(args[0], args[1:]...)


pkg/daemon/daemon.go
```diff
@@ -253,9 +264,28 @@ func (dn *Daemon) addProfileConfig(socketPath string, configFile string, nodePro
 	return nil
 }
 
+// Add fifo scheduling if specified in nodeProfile
+func addScheduling(nodeProfile *ptpv1.PtpProfile, cmdLine string) string {
+	if nodeProfile.PtpSchedulingPolicy != nil && *nodeProfile.PtpSchedulingPolicy == "SCHED_FIFO" {
+		if nodeProfile.PtpSchedulingPriority == nil {
+			glog.Errorf("Priority must be set for SCHED_FIFO; using default scheduling.")
+			return cmdLine
```
SchSeba, line 272: should we start the process if the configuration is not right?

It will be super hard for the user to understand that is not running with the policy

pkg/daemon/daemon.go
```diff
@@ -253,9 +264,28 @@ func (dn *Daemon) addProfileConfig(socketPath string, configFile string, nodePro
...
+		}
+		priority := *nodeProfile.PtpSchedulingPriority
+		if priority < 1 || priority > 65 {
+			glog.Errorf("Invalid priority %d; using default scheduling.", priority)
+			return cmdLine
```
SchSeba, line 277: same as above

pkg/daemon/daemon.go
```diff
@@ -253,9 +264,28 @@ func (dn *Daemon) addProfileConfig(socketPath string, configFile string, nodePro
...
 }
 
+// Add fifo scheduling if specified in nodeProfile
+func addScheduling(nodeProfile *ptpv1.PtpProfile, cmdLine string) string {
+	if nodeProfile.PtpSchedulingPolicy != nil && *nodeProfile.PtpSchedulingPolicy == "SCHED_FIFO" {
+		if nodeProfile.PtpSchedulingPriority == nil {
+			glog.Errorf("Priority must be set for SCHED_FIFO; using default scheduling.")
+			return cmdLine
+		}
+		priority := *nodeProfile.PtpSchedulingPriority
```
SchSeba, line 274: before you do this please check if the variable is not nil.

pkg/daemon/daemon.go
```diff
@@ -253,9 +264,28 @@ func (dn *Daemon) addProfileConfig(socketPath string, configFile string, nodePro
 	return nil
 }
 
+// Add fifo scheduling if specified in nodeProfile
+func addScheduling(nodeProfile *ptpv1.PtpProfile, cmdLine string) string {
+	if nodeProfile.PtpSchedulingPolicy != nil && *nodeProfile.PtpSchedulingPolicy == "SCHED_FIFO" {
```
SchSeba, line 269: why we add the PtpSchedulingPolicy if the only policy here is `SCHED_FIFO` ?

SchSeba (commented):
thanks for the PR!

I leave some comments

pkg/daemon/daemon.go
josephdrichard: The intent is to block this from being configured in the ptp operator, so this check is basically redundant.  A user would need to somehow disable validation once it's in.  In my opinion, logging if this somehow happens is enough.

pkg/daemon/daemon.go
josephdrichard: This is done on line 274.

josephdrichard: Done on line 270

pkg/daemon/daemon.go
josephdrichard: If SCHED_OTHER, then no need to change it as that's the default.

SchSeba: there are move types

```
 -b, --batch          set policy to SCHED_BATCH
 -d, --deadline       set policy to SCHED_DEADLINE
 -f, --fifo           set policy to SCHED_FIFO
 -i, --idle           set policy to SCHED_IDLE
 -o, --other          set policy to SCHED_OTHER
 -r, --rr             set policy to SCHED_RR (default)
```

If the only one we support so just lets remove this `PtpSchedulingPolicy` from the API

pkg/daemon/daemon.go
SchSeba: I don't think this is right let's say there is a request to have the option to disable the webhook (it will happen) then for the user the linuxptp will start but he will see performance issues on open bugs on that.

It will take much more time to debug a case to find the chrt was not applied than just not start the ptp4l process with the error why.

remember we print all the ptp4l outputs it will be really hard to find this (maybe the log will get rotated)

pkg/daemon/daemon.go
```diff
@@ -253,9 +264,28 @@ func (dn *Daemon) addProfileConfig(socketPath string, configFile string, nodePro
...
+			glog.Errorf("Priority must be set for SCHED_FIFO; using default scheduling.")
+			return cmdLine
+		}
+		priority := *nodeProfile.PtpSchedulingPriority
+		if priority < 1 || priority > 65 {
+			glog.Errorf("Invalid priority %d; using default scheduling.", priority)
+			return cmdLine
+		}
+		return fmt.Sprintf("/bin/chrt -f %d %s", priority, cmdLine)
+	}
```
SchSeba, line 280: also there is no validation if `PtpSchedulingPriority != nil` but `PtpSchedulingPolicy == nil` no error of message to the user

josephdrichard: In my opinion setting PtpSchedulingPriority for a PtpSchedulingPolicy other than SCHED_FIFO should just be ignored.
That said, if it is necessary to block, it should be done in the ptp operator.

That said, I don't think we really need to handle the case of invalid input.  I think it is reasonable that if a user both disables the validation, and uses invalid input, then they need to enter correct input or deal with the consequences.  The reason we added the webhook was to handle this.

With the current implementation of linuxptp-daemon, we don't have good error handling, and don't handle ptp4l not starting.  I don't think it is appropriate to add another error path until we have proper error handling.  I'm open to improving error handling in ptp4l, but that's separate from this.
Also, from a user perspective, no harm will come if the system uses default scheduling; it just may not be as accurate as with fifo scheduling, but is still better than ptp4l not running at all.

SchSeba: if this is the path you want to go it's fine from my side

pkg/daemon/daemon.go
```diff
@@ -253,9 +264,28 @@ func (dn *Daemon) addProfileConfig(socketPath string, configFile string, nodePro
...
+	return cmdLine
```
SchSeba, line 281: do you want to add a glog.error or warning here? to print out that the scheduler didn't change?

pkg/daemon/daemon.go
```diff
@@ -253,9 +264,28 @@ func (dn *Daemon) addProfileConfig(socketPath string, configFile string, nodePro
...
+		if nodeProfile.PtpSchedulingPriority == nil {
+			glog.Errorf("Priority must be set for SCHED_FIFO; using default scheduling.")
+			return cmdLine
+		}
+		priority := *nodeProfile.PtpSchedulingPriority
+		if priority < 1 || priority > 65 {
+			glog.Errorf("Invalid priority %d; using default scheduling.", priority)
+			return cmdLine
+		}
+		return fmt.Sprintf("/bin/chrt -f %d %s", priority, cmdLine)
```
SchSeba, line 279: may a log.info here to tell we change the default scheduler for process <process-name>

pkg/daemon/daemon.go
josephdrichard: I don't think it's necessary.  This is the normal use-case where a user does not specify FIFO, so it will use default scheduling.  My intent here was for this change to be invisible to someone who doesn't specify fifo

pkg/daemon/daemon.go
josephdrichard: okay I will make that change

PR #64, opened 2021-12-15
Bug 2016185: Add restart for ptp4l and phc2sys

josephdrichard:
If ptp4l or phc2sys processes die, they should be restarted by
linuxptp-daemon.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -22,6 +23,7 @@ const (
 	PTP4L_CONF_FILE_PATH    = "/etc/ptp4l.conf"
 	PTP4L_CONF_DIR          = "/ptp4l-conf"
 	connectionRetryInterval = 1 * time.Second
+	processRestartInterval  = 1 * time.Second
 	eventSocket             = "/cloud-native/events.sock"
 )
 
@@ -40,6 +42,8 @@ type ptpProcess struct {
 	ptp4lConfigPath string
 	configName      string
 	exitCh          chan bool
+	execMutex       sync.Mutex
+	stopped         bool
 	cmd             *exec.Cmd
 }
 
@@ -195,6 +199,7 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile) 
 			ifaces:     strings.Split(*nodeProfile.Interface, ","),
 			configName: configFile,
 			exitCh:     make(chan bool),
+			stopped:    false,
 			cmd:        phc2sysCreateCmd(nodeProfile)})
 	} else {
 		glog.Infof("applyNodePtpProfile: not starting phc2sys, phc2sysOpts is empty")
@@ -213,6 +218,7 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile) 
 		ptp4lSocketPath: socketPath,
 		configName:      configFile,
 		exitCh:          make(chan bool),
+		stopped:         false,
 		cmd:             ptp4lCreateCmd(nodeProfile, configPath)})
 
 	return nil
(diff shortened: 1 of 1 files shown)


josephdrichard:
@aneeshkp

pkg/daemon/daemon.go
```diff
@@ -40,6 +41,8 @@ type ptpProcess struct {
 	ptp4lConfigPath string
 	configName      string
 	exitCh          chan bool
+	restartMutex    sync.Mutex
```
aneeshkp, line 44: why do you need  Mutex?, Looks like each process is run by its own goroutine

pkg/daemon/daemon.go
```diff
@@ -317,72 +319,92 @@ func cmdRun(p *ptpProcess, stdoutToSocket bool) {
...
 				done <- struct{}{}
-			default:
-				c, err = net.Dial("unix", eventSocket)
-				if err != nil {
-					glog.Errorf("error trying to connect to event socket")
-					time.Sleep(connectionRetryInterval)
-					goto connect
+			}()
+		} else {
+			go func() {
```
aneeshkp, line 347: Memory leak: This goroutine should be released  when recreate is called

pkg/daemon/daemon.go
```diff
@@ -317,72 +319,92 @@ func cmdRun(p *ptpProcess, stdoutToSocket bool) {
...
+			connect:
```
aneeshkp, line 348: Memory leak: This goroutine should be released  when recreate is called

pkg/daemon/daemon.go
```diff
@@ -317,72 +319,92 @@ func cmdRun(p *ptpProcess, stdoutToSocket bool) {
...
+		p.cmd.Stderr = os.Stderr
+		cmdReader, err := p.cmd.StdoutPipe()
+		if err != nil {
+			glog.Errorf("cmdRun() error creating StdoutPipe for %s: %v", p.name, err)
+			return
+		}
+		if !stdoutToSocket {
+			scanner := bufio.NewScanner(cmdReader)
+			go func() {
+				for scanner.Scan() {
```
aneeshkp, line 339: Here too, on loop this go routine need to close?

pkg/daemon/daemon.go
```diff
@@ -317,72 +319,92 @@ func cmdRun(p *ptpProcess, stdoutToSocket bool) {
...
+					_, err := c.Write([]byte(out))
+					if err != nil {
+						glog.Errorf("Write error:", err)
+						goto connect
+					}
 				}
+				done <- struct{}{}
+			}()
+		}
+		p.restartMutex.Lock() // Don't restart after termination
```
aneeshkp, line 374: Not required, It will fall through to recreate label(which you can remove)

pkg/daemon/daemon.go
```diff
@@ -317,72 +319,92 @@ func cmdRun(p *ptpProcess, stdoutToSocket bool) {
...
+		restart := p.restart
```
aneeshkp, line 375: Not required, It will fall through to recreate label(which you can remove)

pkg/daemon/daemon.go
```diff
@@ -317,72 +319,92 @@ func cmdRun(p *ptpProcess, stdoutToSocket bool) {
...
+		if restart {
```
aneeshkp, line 376: remove if condition, it has to go through once

pkg/daemon/daemon.go
```diff
@@ -317,72 +319,92 @@ func cmdRun(p *ptpProcess, stdoutToSocket bool) {
...
-	}
-
-	err = p.cmd.Start()
-	if err != nil {
-		glog.Errorf("cmdRun() error starting %s: %v", p.name, err)
-		return
-	}
-
-	<-done
+		} else {
```
aneeshkp, line 383: no need else condition

pkg/daemon/daemon.go
```diff
@@ -317,72 +319,92 @@ func cmdRun(p *ptpProcess, stdoutToSocket bool) {
...
+			}()
+		}
+		p.restartMutex.Lock() // Don't restart after termination
+		restart := p.restart
+		if restart {
+			err = p.cmd.Start()
+			p.restartMutex.Unlock()
+			if err != nil {
+				glog.Errorf("cmdRun() error starting %s: %v", p.name, err)
+				goto recreateCmd
```
aneeshkp, line 381: don't need goto, end of this condition will take it to recreate block

pkg/daemon/daemon.go
```diff
@@ -317,72 +319,92 @@ func cmdRun(p *ptpProcess, stdoutToSocket bool) {
...
+		} else {
+			p.restartMutex.Unlock()
+		}
 
-	err = p.cmd.Wait()
-	if err != nil {
-		glog.Errorf("cmdRun() error waiting for %s: %v", p.name, err)
-		return
+		<-done
+		if restart {
```
aneeshkp, line 388: Don;'t ned restart check here

pkg/daemon/daemon.go
```diff
@@ -317,72 +319,92 @@ func cmdRun(p *ptpProcess, stdoutToSocket bool) {
...
+			err = p.cmd.Wait()
+			if err != nil {
+				glog.Errorf("cmdRun() error waiting for %s: %v", p.name, err)
+			}
+		}
+	recreateCmd: // Recreate command after it terminates.
```
aneeshkp, line 394: can remove recreateCmd lable

aneeshkp (changes requested):
Please review the comments

pkg/daemon/daemon.go
josephdrichard: restart is originally true, so it will go through.

pkg/daemon/daemon.go
josephdrichard: Was there to skip waiting if error.  Switching to a bool

pkg/daemon/daemon.go
aneeshkp: yes, it is not needed here. The recreate logic only needs to check if it has to recreate.

pkg/daemon/daemon.go
aneeshkp: Previously it was returning on error since now you don't want  to return here and check for recreating if you don't need that return and don't need goto label

pkg/daemon/daemon.go
```diff
@@ -303,86 +308,108 @@ func ptp4lCreateCmd(nodeProfile *ptpv1.PtpProfile, confFilePath string) *exec.Cm
...
+			go func() {
+				for scanner.Scan() {
+					output := scanner.Text()
+					fmt.Printf("%s\n", output)
+					extractMetrics(p.configName, p.name, p.ifaces, output)
+				}
+				done <- struct{}{}
+			}()
+		} else {
+			go func() {
```
aneeshkp, line 339: every time you go over a restart, won't this spawn a  new go routine and the old one will still be running and trying to reconnect?

pkg/daemon/daemon.go
```diff
@@ -303,86 +308,108 @@ func ptp4lCreateCmd(nodeProfile *ptpv1.PtpProfile, confFilePath string) *exec.Cm
...
+			if err != nil {
+				glog.Errorf("cmdRun() error waiting for %s: %v", p.name, err)
+			}
+		}
 
-	err = p.cmd.Wait()
-	if err != nil {
-		glog.Errorf("cmdRun() error waiting for %s: %v", p.name, err)
-		return
+		time.Sleep(5 * time.Second) // Delay to prevent flooding restarts if startup fails
```
aneeshkp, line 393: Move this to a variable. also, 5 sec is too long between restarts. I would suggest 1 sec

pkg/daemon/daemon.go
```diff
@@ -303,86 +308,108 @@ func ptp4lCreateCmd(nodeProfile *ptpv1.PtpProfile, confFilePath string) *exec.Cm
...
 
-	err = p.cmd.Start()
-	if err != nil {
-		glog.Errorf("cmdRun() error starting %s: %v", p.name, err)
-		return
-	}
+		<-done // goroutine is done
 
-	<-done
+		if restart && !skipWait {
```
aneeshkp, line 386: Not sure, why do we need this if condition?

josephdrichard: If there's an error starting, no point waiting.

pkg/daemon/daemon.go
josephdrichard: Okay, I can cut it to one second.  Okay, I can move the time to a constant.

pkg/daemon/daemon.go
```diff
@@ -303,86 +309,98 @@ func ptp4lCreateCmd(nodeProfile *ptpv1.PtpProfile, confFilePath string) *exec.Cm
...
+					_, err := c.Write([]byte(out))
+					if err != nil {
+						glog.Errorf("Write error:", err)
+						goto connect
+					}
 				}
+				done <- struct{}{}
+			}()
+		}
+		p.execMutex.Lock() // Don't restart after termination
```
nishant-parekh, line 367: I dont think we need this mutex lock here only need it from L385

josephdrichard: The case it's here for is if ptp4l dies, and the command is recreated at 392, then it loops back here, and when cmdRun is at this line, cmdStop is called at the same time. I don't want it to send the stop to the process before starting it, and then start it anyways. If that race condition happens, even though it is incredibly unlikely, it could leave behind the process to never (until the node is restarted) get cleaned up.

aneeshkp (approved):
Approving it, but I see there is a chance of a false error message with current changes.

PR #72, opened 2022-02-08
Bug 2054351:Fix restart process  that was holding  lock on asynchronous call, avoid multiple socket error 

aneeshkp:
Signed-off-by: Aneesh Puttur <aneeshputtur@gmail.com>
Cleaning termination with an exit in defer function and not holding the lock for asynchronous calls like cmd start and sig term.
This was causing the listener at the sidecar to see multiple socket connections (2 per termination)

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -46,6 +46,19 @@ type ptpProcess struct {
 	cmd             *exec.Cmd
 }
 
+func (p ptpProcess) Stopped() bool {
+	p.execMutex.Lock()
+	me := p.stopped
+	p.execMutex.Unlock()
+	return me
+}
+
+func (p *ptpProcess) setStopped(val bool) {
+	p.execMutex.Lock()
+	p.stopped = val
+	p.execMutex.Unlock()
+}
+
 // Daemon is the main structure for linuxptp instance.
 // It contains all the necessary data to run linuxptp instance.
 type Daemon struct {
@@ -67,7 +80,7 @@ type Daemon struct {
 	stopCh <-chan struct{}
 }
 
-// NewLinuxPTP is called by daemon to generate new linuxptp instance
+// New LinuxPTP is called by daemon to generate new linuxptp instance
 func New(
 	nodeName string,
 	namespace string,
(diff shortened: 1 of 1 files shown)


pkg/daemon/daemon.go
```diff
@@ -363,43 +384,35 @@ func cmdRun(p *ptpProcess, stdoutToSocket bool) {
...
-
 		err = p.cmd.Wait()
 		if err != nil {
 			glog.Errorf("cmdRun() error waiting for %s: %v", p.name, err)
 		}
 
 		time.Sleep(connectionRetryInterval) // Delay to prevent flooding restarts if startup fails
-		p.execMutex.Lock()                  // Don't restart after termination
+		// Don't restart after termination
 		if p.stopped {
```
josephdrichard, line 402: p.Stopped()

pkg/daemon/daemon.go
```diff
@@ -312,6 +325,14 @@ func ptp4lCreateCmd(nodeProfile *ptpv1.PtpProfile, confFilePath string) *exec.Cm
 func cmdRun(p *ptpProcess, stdoutToSocket bool) {
 	var c net.Conn
 	done := make(chan struct{}) // Done setting up logging.  Go ahead and wait for process
+	defer func() {
```
josephdrichard, line 328: Is everything outside of this defer func just style changes, or necessary for this fix?
If it's not part of this fix, can you put it in a separate commit?

josephdrichard (commented):
Is there an open bug for this?

pkg/daemon/daemon.go
aneeshkp: This is part of the restart process and I want to backport this once it is thoroughly tested by QE.
So it is good to keep it in one PR to track the fix or for any issues to reverse the changes.

josephdrichard: Can you split it into two separate commits within the same PR?

josephdrichard:
Commit message seems to have been incorrectly changed.

pkg/daemon/daemon.go
aneeshkp: I don't get it , what two PR?

aneeshkp:
> > Is there an open bug for this?
> 
> @josephdrichard , Haven't created a bug for this. I spoke to Ken and told him we will create a bug if we backport only, that is after successful testing. So as of now, no plans to do any backport PR immediately.

created bug and updated

aneeshkp:
> Commit message seems to have been incorrectly changed.

amended the commit message , Thanks

PR #73, opened 2022-02-09
Remove adding flags for monitor

josephdrichard:


pkg/daemon/metrics.go
```diff
@@ -416,37 +414,6 @@ func extractPTP4lEventState(output string) (portId int, role ptpPortRole) {
 	return
 }
 
-func addFlagsForMonitor(nodeProfile *ptpv1.PtpProfile, stdoutToSocket bool) {
```
josephdrichard, line 419: Is this still something we want?  If so, it should be added by operator not here.

aneeshkp: This change might  break cnf-test 
Also - m  required for regular Prometheus metrics (for non-events,)  and  -u to be removed for events to generated S1/S2 clock state.
Unless you add these as part of your webhook validations for ptpconfig and ptpoperatorconfig (foe event flag) updates

aneeshkp: How can an operator add this flag? are you suggesting modifying ptpconfig CR  by the operator when the event is enabled and disabled?
I am okay with these changes only If we can retain this functionality somewhere else and then remove this. Also please run and functests  test for these changes.

josephdrichard: Is there any reason we can't just expect a user to provision this correctly?

pkg/daemon/config.go
```diff
@@ -157,3 +157,37 @@ func (conf *ptp4lConf) renderPtp4lConf() (string, string) {
 	}
 	return configOut, strings.Join(conf.mapping, ",")
 }
+
+func addFlagsForMonitor(nodeProfile *ptpv1.PtpProfile, conf *ptp4lConf, stdoutToSocket bool) {
```
aneeshkp, line 161: Only required change in this PR looks like 
_, exist := conf.sections["[global]"].options["summary_interval"]
Why changing so many code?

josephdrichard:
Moved change back to metric file instead of config file.
/hold While I retest

josephdrichard:
Testing looks good.
/hold cancel

PR #78, opened 2022-04-08
Generate ptp4l/phc2sys process status metrics 

aneeshkp:
Generate ptp4l/phc2sys process status  metrics  and report  status to sidecar when event  framework is  enabled

Adding health check metrics for the PTP4l process and Phc2sys process.
Generate metrics for process status and send ptp4l status and pch2sys status to event framework.

Signed-off-by: Aneesh Puttur <aneeshputtur@gmail.com>

### DIFF
--- a/pkg/daemon/metrics.go
+++ b/pkg/daemon/metrics.go
@@ -2,13 +2,14 @@ package daemon
 
 import (
 	"fmt"
-	"github.com/golang/glog"
-	"github.com/prometheus/client_golang/prometheus/collectors"
 	"net/http"
 	"strconv"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/golang/glog"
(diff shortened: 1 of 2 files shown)


pkg/daemon/metrics.go
```diff
@@ -367,6 +385,17 @@ func UpdateInterfaceRoleMetrics(process string, iface string, role ptpPortRole)
 		"process": process, "node": NodeName, "iface": iface}).Set(float64(role))
 }
 
+func UpdateProcessStatusMetrics(process string, cfgName string, status int64) {
```
josephdrichard, line 388: Can you define "up" and "down" constants that you pass to UpdateProcessStatusMetrics as the status instead of an int64?

aneeshkp: Done

josephdrichard (approved):
/lgtm
/approved

PR #83, opened 2022-07-29
Bug 2106476: Preserve iface ordering in ptpconfig

josephdrichard:


pkg/daemon/config.go
```diff
@@ -107,35 +108,42 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
...
+				globalIsDefined = true
+			}
+			currentSection = ptp4lConfSection{options: map[string]string{}, sectionName: currentSectionName}
+		} else if currentSectionName != "" {
 			split := strings.IndexByte(line, ' ')
 			if split > 0 {
-				section := output.sections[currentSection]
-				section.options[line[:split]] = line[split:]
-				output.sections[currentSection] = section
+				currentSection.options[line[:split]] = line[split:]
```
aneeshkp, line 136: So the section are added to an array, that way section is guaranteed to be in order but the content are not, correct ?

josephdrichard: correct.  There is no requirement for the order within a section to be kept.

aneeshkp (approved):

PR #105, opened 2022-12-04
OCPBUGS-2806: Add log filtering

josephdrichard:
/hold
Need to update and retest after https://github.com/openshift/ptp-operator/pull/277 merges so I can point to correct ptp operator revision.


josephdrichard:
Testing looks good.
/hold cancel

aneeshkp (approved):

PR #110, opened 2023-01-08
Add support for ts2phc in linuxptp-daemon

josephdrichard:
/hold
still need more testing as well as vendoring in https://github.com/openshift/ptp-operator/pull/287 .  Pushing first draft for review.

### DIFF
--- a/pkg/daemon/config.go
+++ b/pkg/daemon/config.go
@@ -106,44 +106,45 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
 	return []ptpv1.PtpProfile{*ptpConfig}, true
 }
 
+// Takes as input a PtpProfile.Ptp4lConf and outputs as ptp4lConf struct
 func (output *ptp4lConf) populatePtp4lConf(config *string) error {
-	lines := strings.Split(*config, "\n")
 	var currentSectionName string
 	var currentSection ptp4lConfSection
 	output.sections = make([]ptp4lConfSection, 0)
 	globalIsDefined := false
 
-	for _, line := range lines {
-		line = strings.TrimSpace(line)
-		if strings.HasPrefix(line, "#") {
-			continue
-		} else if strings.HasPrefix(line, "[") {
-			if currentSectionName != "" {
-				output.sections = append(output.sections, currentSection)
-			}
-			currentSectionName = line
-			currentLine := strings.Split(line, "]")
+	if config != nil {
+		for _, line := range strings.Split(*config, "\n") {
+			line = strings.TrimSpace(line)
+			if strings.HasPrefix(line, "#") {
+				continue
+			} else if strings.HasPrefix(line, "[") {
+				if currentSectionName != "" {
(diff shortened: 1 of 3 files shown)


pkg/daemon/config.go
```diff
@@ -107,43 +107,45 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
...
-	for _, line := range lines {
-		line = strings.TrimSpace(line)
-		if strings.HasPrefix(line, "#") {
-			continue
-		} else if strings.HasPrefix(line, "[") {
-			if currentSectionName != "" {
-				output.sections = append(output.sections, currentSection)
+	if config != nil {
+		for _, line := range strings.Split(*config, "\n") {
+			glog.Infof(line)
```
jzding, line 117: Is this log temporary? If not then need to add a context log explaining what this is.

josephdrichard: Not needed.  will remove.

pkg/daemon/config.go
```diff
@@ -107,43 +107,45 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
 }
 
 func (output *ptp4lConf) populatePtp4lConf(config *string) error {
```
jzding, line 109: it will be helpful to add a sample config string as comments of the function

pkg/daemon/config.go
```diff
@@ -107,43 +107,45 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
...
+		for _, line := range strings.Split(*config, "\n") {
+			glog.Infof(line)
+			line = strings.TrimSpace(line)
+			if strings.HasPrefix(line, "#") {
+				continue
+			} else if strings.HasPrefix(line, "[") {
+				if currentSectionName != "" {
+					output.sections = append(output.sections, currentSection)
+				}
+				currentSectionName = line
```
jzding, line 125: this line is useless

pkg/daemon/config.go
```diff
@@ -107,43 +107,45 @@ func tryToLoadOldConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
...
+				currentLine := strings.Split(line, "]")
+
+				if len(currentLine) < 2 {
+					return errors.New("Section missing closing ']'")
```
jzding, line 129: include value of `currentLine` would help debug

josephdrichard: updated

pkg/daemon/config.go
josephdrichard: removed

pkg/daemon/config.go
josephdrichard: Added a comment about what it does.  I don't think a sample config string would fit well here because of how large it is, and means that we need to keep it updated whenever we make changes to configs.

jzding (approved):
/lgtm

PR #117, opened 2023-02-02
Fix phc2sys metrics

josephdrichard:


pkg/daemon/daemon.go
```diff
@@ -258,7 +258,7 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 			socketPath = fmt.Sprintf("/var/run/ptp4l.%d.socket", runID)
 			configFile = fmt.Sprintf("phc2sys.%d.config", runID)
 			configPath = fmt.Sprintf("/var/run/%s", configFile)
-			messageTag = fmt.Sprintf("[ptp4l.%d.config]", runID)
+			messageTag = fmt.Sprintf("[phc2sys.%d.config]", runID)
```
aneeshkp, line 261: Do we create multiple config now ?

aneeshkp: phc2sys.0.config, ptp4l.0.config & ts2phc.0.config ?

aneeshkp: can we have it the way it was before 
phc2sys[861009.766]: [ptp4l.0.config] ens3f1 rms    3 max    3 freq  -3004 +/-   0 delay  2523 +/-   0
ptp4l[861009.799]: [ptp4l.0.config] master offset          2 s2 freq   -3004 path delay       357

PR #125, opened 2023-02-17
OCPBUGS-8249: add support for multinode service for http moving from 4.13 to 4.12

aneeshkp:
This PR adds  first part of nodename as label to daemon pod. Which is needed for http  events service to work on multi node.
http is not supported  in 4.13 but need to back port to 4.12
Need to back port  https://github.com/openshift/ptp-operator/pull/282 to support multi node.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -1,12 +1,15 @@
 package main
 
 import (
+	"context"
+	"encoding/json"
 	"flag"
 	"io/ioutil"
 	"os"
 	"os/signal"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cmd/main.go
```diff
@@ -135,3 +144,35 @@ func main() {
...
+	Value string `json:"value"`
+}
+
+func labelPod(kubeClient *kubernetes.Clientset, nodeName, podName string) {
+	pod, err := kubeClient.CoreV1().Pods(daemon.PtpNamespace).Get(context.TODO(), podName, metav1.GetOptions{})
+	if err != nil {
+		glog.Error(err)
+	}
+	if pod == nil {
+		glog.Info("Could not find linux-ptp-daemon pod to label")
```
josephdrichard, line 160: linuxptp-daemon

PR #142, opened 2023-05-30
CNF-8665: Update GM ClockClass and ClockQuality based on events

jzding:


pkg/event/event.go
```diff
@@ -152,52 +159,63 @@ func (e *EventHandler) ProcessEvents() {
...
 		}
-	case PTP_HOLDOVER:
+	case PTP_FREERUN, GNSS_FREERUN:
 		switch clockType {
 		case GM:
-			changeClockType(cfgName, pmc.CmdUpdateGMClass_LOCKED, 248)
+			cq.ClockClass = ClockClassFreerun
+			cq.ClockAccuracy = protocol.ClockAccuracyUnknown
+			cq.OffsetScaledLogVariance = 0xffff
+			runUpdateGMClockClass(cfgName, cq)
```
aneeshkp, line 192: do you want to add for OC and BC

jzding: we do GM only for now

PR #143, opened 2023-06-03
Add ublx config to plugin

josephdrichard:
Changes to plugin structure also required to allow sharing data between OnPTPConfigChange and PopulateHwConfig, which is required to write the status to the nodeptpdevice

### DIFF
--- a/addons/intel/e810.go
+++ b/addons/intel/e810.go
@@ -2,14 +2,26 @@ package intel
 
 import (
 	"encoding/json"
+	"fmt"
 	"github.com/golang/glog"
 	"github.com/openshift/linuxptp-daemon/pkg/plugin"
 	ptpv1 "github.com/openshift/ptp-operator/api/v1"
 	"os/exec"
+	"strings"
 )
 
 type E810Opts struct {
-	EnableDefaultConfig bool `json:"enableDefaultConfig"`
+	EnableDefaultConfig bool           `json:"enableDefaultConfig"`
+	UblxCmds            []E810UblxCmds `json:"ublxCmds"`
+}
+
+type E810UblxCmds struct {
+	ReportOutput bool     `json:"reportOutput"`
+	Args         []string `json:"args"`
+}
+
+type E810PluginData struct {
+	hwplugins *[]string
 }
 
 // Sourced from https://github.com/RHsyseng/oot-ice/blob/main/ptp-config.sh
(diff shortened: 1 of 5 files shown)


addons/intel/e810.go
```diff
@@ -56,21 +68,87 @@ func OnPTPConfigChangeE810(nodeProfile *ptpv1.PtpProfile) error {
 	return nil
 }
 
-func PopulateHwConfigE810(hwconfigs *[]ptpv1.HwConfig) error {
+func AfterRunPTPCommandE810(data *interface{}, nodeProfile *ptpv1.PtpProfile, command string) error {
+	glog.Info("calling onPTPConfigChange for e810 plugin")
```
jzding, line 72: calling AfterRunPTPCommandE810

addons/intel/e810.go
```diff
@@ -56,21 +68,87 @@ func OnPTPConfigChangeE810(nodeProfile *ptpv1.PtpProfile) error {
...
+	if data != nil {
+		_data := *data
+		var pluginData *E810PluginData = _data.(*E810PluginData)
+		_pluginData := *pluginData
+		if _pluginData.hwplugins != nil {
+			for _, _hwconfig := range *_pluginData.hwplugins {
+				hwConfig := ptpv1.HwConfig{}
+				hwConfig.DeviceID = "e810"
+				hwConfig.Status = _hwconfig
+				*hwconfigs = append(*hwconfigs, hwConfig)
```
jzding, line 132: if there is a hwconfig already exist with DeviceID=="e810", do we replace it or add another resulting in multiple hwconfigs with same DeviceID?

pkg/daemon/daemon.go
```diff
@@ -233,15 +234,18 @@ func (dn *Daemon) applyNodePTPProfiles() error {
...
-						time.Sleep(2 * time.Second)
+						time.Sleep(5 * time.Second)
+						dn.pluginManager.AfterRunPTPCommand(p.nodeProfile, d.Name())
 						d.monitorEvent(clockType, p.configName, p.exitCh, dn.processManager.eventChannel)
 					}
 				}
 			}
 			time.Sleep(1 * time.Second)
 			go p.cmdRun(dn.stdoutToSocket)
+			dn.pluginManager.AfterRunPTPCommand(p.nodeProfile, p.name)
```
jzding, line 245: do we need to add a few second delay before running this command?

addons/intel/e810.go
josephdrichard: Fixed

pkg/daemon/daemon.go
josephdrichard: We shouldn't.  If necessary for any specific plugin, this can be added inside of the plugin that requires it.

josephdrichard:
submitted-for: CNF-7017

PR #144, opened 2023-06-22
OCPBUGS-14681: fix clock class updates

vitus133:
This commit introduces periodical clock class checks to indicate when there is a change

/cc @aneeshkp @josephdrichard @jzding @nishant-parekh

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -47,19 +47,21 @@ type ProcessManager struct {
 }
 
 type ptpProcess struct {
-	name            string
-	ifaces          []string
-	ptp4lSocketPath string
-	ptp4lConfigPath string
-	configName      string
-	messageTag      string
-	exitCh          chan bool
-	execMutex       sync.Mutex
-	stopped         bool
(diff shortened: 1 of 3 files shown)


pkg/daemon/daemon.go
```diff
@@ -531,39 +581,18 @@ func (p *ptpProcess) cmdRun(stdoutToSocket bool) {
...
+						p.pmcCheck = false
+						go p.updateClockClass(&c)
+					}
+
 					if regexErr != nil || !logFilterRegex.MatchString(output) {
 						fmt.Printf("%s\n", output)
 					}
 					out := fmt.Sprintf("%s\n", output)
 					if p.name == ptp4lProcessName {
 						if strings.Contains(output, ClockClassChangeIndicator) {
```
aneeshkp, line 594: Do we need this check anymore if we are doing clock class check  in interval ?

vitus133: I thought leaving it here for after ptp4l implements class change notification.
WDYT?

PR #145, opened 2023-06-23
Add monitoring for all process and setting up GM State

aneeshkp:
This PR monitors DPLL, GNSS,ts2phc and derives GM state.
Update new metrics to all the components that are monitored by GM monitoring process


pkg/daemon/gpsd.go
```diff
@@ -112,13 +116,89 @@ func (g *gpsd) cmdRun(stdoutToSocket bool) {
...
+				offsetS, err2 := ublx.GetNavOffset()
+				if errs == nil && err2 == nil {
+					//calculate PTP states
+					g.offset, _ = strconv.ParseInt(offsetS, 10, 64)
+					if nStatus > 3 && g.isOffsetInRange() {
+						g.state = event.PTP_LOCKED
+					} else {
+						g.state = event.PTP_FREERUN
+					}
+					processCfg.EventChannel <- event.EventChannel{
```
jzding, line 163: should we save the previous g.state and send out events only if g.state is changed?

pkg/daemon/metrics.go
```diff
@@ -174,7 +174,7 @@ func updatePTPMetrics(from, process, iface string, ptpOffset, maxPtpOffset, freq
 }
 
 // extractMetrics ...
```
jzding, line 176: it would be nice to add a comment with sample metrics

pkg/dpll/dpll.go
```diff
@@ -0,0 +1,268 @@
...
+	"github.com/openshift/linuxptp-daemon/pkg/event"
+)
+
+const (
+	DPLL_UNKNOW        = -1
+	DPLL_INVALID       = 0
+	DPLL_FREERUN       = 1
+	DPLL_LOCKED        = 2
+	DPLL_LOCKED_HO_ACQ = 3
+	DPLL_HOLOVER       = 4
```
jzding, line 21: When we compare phase and frequency states we choose the lower one, which imply the later the number, the worse the signal.  Based on this the values should be the following:
```
	DPLL_HOLOVER       = 2
	DPLL_LOCKED_HO_ACQ = 3
	DPLL_LOCKED        = 4
```

nishant-parekh (approved):

PR #151, opened 2023-06-30
compute clockType from ptp4lConfig

nishant-parekh:
GM = No Slave interface
  BC = Multiple interfaces with atleast one slave
  OC = Single slave interface

### DIFF
--- a/pkg/daemon/config.go
+++ b/pkg/daemon/config.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/openshift/linuxptp-daemon/pkg/event"
 	"io/ioutil"
 	"os"
 	"strings"
@@ -32,6 +33,7 @@ type ptp4lConf struct {
 	sections     []ptp4lConfSection
 	mapping      []string
 	profile_name string
+	clock_type   event.ClockType
 }
 
 func NewLinuxPTPConfUpdate() (*LinuxPTPConfUpdate, error) {
@@ -112,6 +114,7 @@ func (output *ptp4lConf) populatePtp4lConf(config *string) error {
 	var currentSection ptp4lConfSection
 	output.sections = make([]ptp4lConfSection, 0)
 	globalIsDefined := false
+	hasSlaveConfigDefined := false
 
 	if config != nil {
 		for _, line := range strings.Split(*config, "\n") {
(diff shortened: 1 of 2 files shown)


pkg/daemon/config.go
```diff
@@ -146,6 +153,16 @@ func (output *ptp4lConf) populatePtp4lConf(config *string) error {
 			output.sections = append(output.sections, currentSection)
 		}
 	}
+	if !hasSlaveInterfaceDefined {
+		// No Slave Interfaces defined
+		output.clock_type = event.GM
+	} else if len(output.sections) > 1 {
```
jzding, line 159: This is not true if there is a [global] section.
Suggest move this logic after Line#168, so that a [global] section is always present, and then change this line to
`} else if len(output.sections) > 2 {`

pkg/daemon/config.go
```diff
@@ -137,6 +140,10 @@ func (output *ptp4lConf) populatePtp4lConf(config *string) error {
 				split := strings.IndexByte(line, ' ')
 				if split > 0 {
 					currentSection.options[line[:split]] = line[split:]
+					if (line[:split] == "masterOnly" && line[split:] == "0") ||
+						(line[:split] == "slaveOnly" && line[split:] == "1") {
+						hasSlaveInterfaceDefined = true
```
josephdrichard, line 145: slaveOnly is set on a per-config basis, not per-interface, so should rename hasSlaveInterfaceDefined

pkg/daemon/config.go
```diff
@@ -32,6 +33,7 @@ type ptp4lConf struct {
 	sections     []ptp4lConfSection
 	mapping      []string
 	profile_name string
+	clock_type   event.ClockType
```
josephdrichard, line 36: ClockType isn't at all related to events, and this events code may be pulled out when upstreaming.  Can you move it to be defined somewhere else?  Maybe github.com/openshift/linuxptp-daemon/pkg/config

pkg/daemon/config.go
```diff
@@ -137,6 +140,10 @@ func (output *ptp4lConf) populatePtp4lConf(config *string) error {
 				split := strings.IndexByte(line, ' ')
 				if split > 0 {
 					currentSection.options[line[:split]] = line[split:]
+					if (line[:split] == "masterOnly" && line[split:] == "0") ||
```
josephdrichard, line 143: masterOnly or serverOnly

pkg/daemon/config.go
```diff
@@ -137,6 +140,10 @@ func (output *ptp4lConf) populatePtp4lConf(config *string) error {
...
+						(line[:split] == "slaveOnly" && line[split:] == "1") {
```
josephdrichard, line 144: slaveOnly or clientOnly

pkg/daemon/config.go
nishant-parekh: updated `hasSlaveInterfaceDefined` => `hasSlaveConfigDefined`

pkg/daemon/config.go
nishant-parekh: extended check to include `serverOnly` as well

pkg/daemon/config.go
nishant-parekh: extended check to include `clientOnly` as well

pkg/daemon/config.go
nishant-parekh: Currently this is the only place we handle a ptp4l config update event, which is what we are using the compute clockType. Please suggest how we could decouple this logic from the existing implementation, and I will be happy to follow up and update this logic in a follow up PR.

josephdrichard (approved):
/lgtm

PR #152, opened 2023-07-06
BUG OCPBUGS-15584: leap-seconds.list file included as part of linuxptp-daemon container expired on June 28, 2023

aneeshkp:
update leap file to latest


jzding (approved):
/lgtm

PR #153, opened 2023-07-06
Bug OCPBUGS-15370: update Dockerfile.rhel7 with Dockerfile content

jzding:
also update builder from golang-1.19 to golang-1.20 in Dockerfile


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.19-openshift-4.14 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.20-openshift-4.14 AS builder
```
josephdrichard, line 1: is there a rhel-9 base instead now?

nishant-parekh: there is one but I don't think any of the operators have updated them yet.

PR #156, opened 2023-07-10
bug fix with metrics and logs

aneeshkp:
Fixed bugs:
GM Metrics
GM Clock Type

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -35,10 +35,6 @@ const (
 	ClockClassChangeIndicator = "selected best master clock"
 )
 
-var (
-	clockType event.ClockType
-)
-
 // ProcessManager manages a set of ptpProcess
 // which could be ptp4l, phc2sys or timemaster.
 // Processes in ProcessManager will be started
@@ -66,6 +62,7 @@ type ptpProcess struct {
 	nodeProfile       *ptpv1.PtpProfile
 	parentClockClass  float64
 	pmcCheck          bool
+	clockType         event.ClockType
 	ptpClockThreshold *ptpv1.PtpClockThreshold
 }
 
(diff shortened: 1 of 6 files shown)


pkg/daemon/daemon.go
```diff
@@ -588,26 +586,33 @@ func (p *ptpProcess) cmdRun(stdoutToSocket bool) {
...
-							CfgName:     p.configName,
-							IFace:       iface,
-							Values: map[event.ValueType]int64{
-								event.OFFSET: int64(ts2phcsOffset),
-							},
-							ClockType:  clockType,
-							Time:       time.Now().Unix(),
-							WriteToLog: true,
-							Reset:      false,
+						updateClockStateMetrics(p.name, iface, FREERUN)
```
jzding, line 593: this line should move to line 599 in an `else` statement of checking Max/MinOffsetThreshold

pkg/ublox/ublox.go
```diff
@@ -31,7 +31,7 @@ const (
 	// CMD_PROTO_VERSION ...
 	CMD_PROTO_VERSION = " -p MON-VER"
 	// CMD_VOLTAGE_CONTROLER ...
-	CMD_VOLTAGE_CONTROLER = " -v 1 -z CFG-HW-ANT_CFG_VOLTCTRL %d"
+	CMD_VOLTAGE_CONTROLER = " -v 1 -z CFG-HW-ANT_CFG_VOLTCTRL, %d"
```
jzding, line 34: This won't work. There should not be a space after `,`.

josephdrichard: This can go in the ptpconfig?

pkg/daemon/daemon.go
aneeshkp: since we are bringing in max/min threshold values,They should be applied for all configurations

jzding (approved):
/lgtm

PR #158, opened 2023-07-11
Bug OCPBUGS-16041: replace pip3 install in Dockerfile for downstream build

jzding:


jzding:
will replace this with a proper cachito solution.

PR #160, opened 2023-07-13
OCPBUGS-16188:Linuxptp-daemon crashing due to index out of range doing DPLL monitoring

aneeshkp:
fixed sysfs reading content
Set gpsd debug level to default 0
added interface to teh log for dpll /gnss/gmstats and gpsd


josephdrichard (approved):
/lgtm

PR #167, opened 2023-07-24
Dpll netlink

vitus133:
This is a draft PR for review and discussion


pkg/dpll/dpll.go
```diff
@@ -43,6 +50,25 @@ type DpllConfig struct {
 	dependingState         []event.EventSource
 	exitCh                 chan struct{}
 	ticker                 *time.Ticker
+	// DPLL netlink connection pointer. If 'nil', use sysfs
+	conn *nl.Conn
+	// We need to keep latest DPLL status values, since Netlink device
+	// change indications don't contain all the status fields, but
+	// only the changed one(s)
```
vitus133, line 57: Make this comment more clear - there are two dplls, and only one can change at a time. Hence need to store the state

pkg/dpll/dpll.go
```diff
@@ -66,6 +92,12 @@ func (d *DpllConfig) CmdStop() {
 	glog.Infof("Ticker stopped %s", d.Name())
 	close(d.exitCh) // terminate loop
 	glog.Infof("Process %s terminated", d.Name())
+	if d.conn != nil {
+		d.nlCtxCancel()
```
vitus133, line 96: Maybe we need a waitgroup here, because we can close the connection before the goroutine will exit. This will lead to attempt to use a closed connection, error in logs

pkg/dpll/dpll.go
```diff
@@ -66,6 +95,12 @@ func (d *DpllConfig) CmdStop() {
 	glog.Infof("Ticker stopped %s", d.Name())
 	close(d.exitCh) // terminate loop
 	glog.Infof("Process %s terminated", d.Name())
+	if d.conn != nil {
+		// d.nlCtxCancel()
```
vitus133, line 99: TODO: remove the commented out context cancel

pkg/dpll/dpll.go
```diff
@@ -94,33 +129,351 @@ func NewDpll(localMaxHoldoverOffSet, localHoldoverTimeout, maxInSpecOffset int64
...
+			if d.conn == nil {
+				conn, err := nl.Dial(nil)
+				if err != nil {
+					d.conn = nil
+					glog.Info("can't dial to DPLL netlink: ", err)
+					goto checkExit
+
+				} else {
+					d.conn = conn
+					// d.nlCtx, d.nlCtxCancel = context.WithCancel(context.Background())
```
vitus133, line 229: TODO: remove

pkg/dpll/dpll.go
```diff
@@ -94,33 +129,351 @@ func NewDpll(localMaxHoldoverOffSet, localHoldoverTimeout, maxInSpecOffset int64
...
+			if err != nil {
+				glog.Info("dpll monitoring is still running")
+				redial = false
+				d.updateDependingProcessState(processCfg)
+				d.stateDecision(processCfg, holdoverCloseCh)
+			} else {
+				glog.Info("dpll monitoring exited, redial")
+				d.stopDpll()
+				redial = true
+			}
```
vitus133, line 293: This is a mechanism to check whether netlink monitoring function exited on error, and if yes - redial. 
If semaphore with timeout expires, no need to redial, as goroutine is still running.

pkg/dpll/dpll.go
```diff
@@ -94,33 +129,351 @@ func NewDpll(localMaxHoldoverOffSet, localHoldoverTimeout, maxInSpecOffset int64
...
+		}
+	}
+}
+
+// stopDpll stops DPLL monitoring
+func (d *DpllConfig) stopDpll() {
+	// d.nlCtxCancel()
```
vitus133, line 300: TODO: remove

pkg/dpll/dpll.go
```diff
@@ -94,33 +129,351 @@ func NewDpll(localMaxHoldoverOffSet, localHoldoverTimeout, maxInSpecOffset int64
...
+	d.conn.Close()
+	d.conn = nil
+}
+
+// MonitorProcess is initiating monitoring of DPLL associated with a process
 func (d *DpllConfig) MonitorProcess(processCfg config.ProcessConfig) {
+	d.processConfig = processCfg
```
vitus133, line 307: @aneeshkp I thought this line was missing somewhere, because d.processConfig remains uninitialized and used in 
```go
func (d *DpllConfig) isOffsetInRange() bool {
	if d.offset < d.processConfig.GMThreshold.Max && d.offset > d.processConfig.GMThreshold.Min {
		return true
	}
	return false
}
```
This will also allow us to remove passing processConfig as a parameter everywhere.
WDYT?

pkg/dpll/dpll.go
```diff
@@ -94,33 +129,351 @@ func NewDpll(localMaxHoldoverOffSet, localHoldoverTimeout, maxInSpecOffset int64
...
 	go d.MonitorDpll(processCfg)
 }
 
-// MonitorDpll ... monitor dpll events
+// MonitorDpll tries to create netlink connection to DPLL.
+// If netlink can't be dialed for N retries and sysfs directory
+// structure is present, fall back to sysfs.
+// Otherwise, use netlink and never fall back
```
vitus133, line 314: Is this logic OK?
WDYT?

pkg/dpll/dpll.go
```diff
@@ -94,33 +129,351 @@ func NewDpll(localMaxHoldoverOffSet, localHoldoverTimeout, maxInSpecOffset int64
...
+	if d.conn == nil && d.isSysFsPresent() {
+		// fall back to sysfs
+		d.MonitorDpllSysfs(processCfg)
+
+	} else { // reconnect to Netlink forever
+		d.MonitorDpllNetlink(processCfg)
+	}
+}
+
+// initDpllColckId finds clock ID associated with the network interface
```
vitus133, line 338: TODO: ClockId

pkg/event/event.go
```diff
@@ -374,18 +374,29 @@ func (e *EventHandler) GetPTPState(source EventSource, cfgName string) PTPState
...
-						request.ResponseChannel <- v.State
+		request = <-e.statusRequestChannel
+		if m, ok := e.data[request.CfgName]; ok {
+			for _, v := range m {
+				if v.ProcessName == request.Source {
+					// Attempt to respond to the channel, or
+					// timeout if nobody listening
+					select {
+					case request.ResponseChannel <- v.State:
+						goto cont
```
vitus133, line 387: `continue` didn't work here, as it continued the inner loop, while we want to continue the outer loop

pkg/event/event.go
```diff
@@ -487,13 +498,11 @@ func (e *EventHandler) unregisterMetrics(configName string, processName string)
...
-	//responseChannel := make(chan string)
-	statusRequestChannel <- request
-
-	// Wait for and receive the response
-	//response := <-responseChannel
-
+	select {
+	case statusRequestChannel <- request:
+	case <-time.After(200 * time.Millisecond):
+		glog.Warning("dpll: failed to send request to statusRequestChannel")
```
vitus133, line 504: TODO: remove `dpll` in the warning, replace by event

vitus133 (commented):
@aneeshkp @jzding @josephdrichard could you guys give it an offline review? I need some answers w.r.t. maximum and minimum offset thresholds

aneeshkp (approved):
/lgtm

PR #168, opened 2023-07-27
Fix ts2phc.nmea_serialport port to be read from ptpconfig for gpsd

aneeshkp:


pkg/daemon/config.go
```diff
@@ -30,10 +31,11 @@ type ptp4lConfSection struct {
...
 type ptp4lConf struct {
-	sections     []ptp4lConfSection
-	mapping      []string
-	profile_name string
-	clock_type   event.ClockType
+	sections       []ptp4lConfSection
+	mapping        []string
+	profile_name   string
+	clock_type     event.ClockType
+	gnssSerialPort string // gnss serial port
```
josephdrichard, line 38: Can you keep underscore instead of camelCase to be consistent with other items in ptp4lConf?

josephdrichard: or you can change the other 2 to use camelCase too

pkg/daemon/daemon.go
```diff
@@ -431,13 +435,19 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
...
-				name:        GPSD_PROCESSNAME,
-				execMutex:   sync.Mutex{},
-				cmd:         nil,
-				serialPort:  GPSD_SERIALPORT,
+				name:      GPSD_PROCESSNAME,
+				execMutex: sync.Mutex{},
+				cmd:       nil,
+				serialPort: func() string {
+					if output.gnssSerialPort == "" {
+						return GPSDDefaultGNSSSerialPort
```
josephdrichard, line 443: I would just set output.gnssSerialPort to the default when initializing the ptp4lConf right before the call to populatePtp4lConf, and then it would be overridden if set.  No need to check here then.

pkg/daemon/config.go
aneeshkp: done

pkg/daemon/daemon.go

pkg/dpll/dpll.go
```diff
@@ -287,7 +287,7 @@ func (d *DpllConfig) sysfs(iface string) (phaseState, frequencyState, phaseOffse
 	// Read the content of the sysfs path
 	content, err = os.ReadFile(frequencyStateStr)
 	if err != nil {
-		glog.Errorf("error reading sysfs path %s %s:", frequencyStateStr, err)
+		glog.Errorf("error reading vi ~sysfs path %s %s:", frequencyStateStr, err)
```
josephdrichard, line 290: ?

aneeshkp: done

Dockerfile
```diff
@@ -29,9 +29,6 @@ RUN yum -y install python3-pip
 
 COPY --from=buildgps /usr/local/lib/python3.6/site-packages /usr/local/lib/python3.6/site-packages
 
-#add ubxtool dependency when gpsd is not used
-RUN pip3 install pyserial
```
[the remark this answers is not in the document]
aneeshkp: yes  we don't need pyserial

josephdrichard (approved):
/lgtm
/approved

Dockerfile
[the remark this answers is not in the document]
jzding: if that's the case you need to remove it from 
[requirements.in ](https://github.com/openshift/linuxptp-daemon/blob/master/requirements.in)
[requirements.txt](https://github.com/openshift/linuxptp-daemon/blob/master/requirements.txt)

pkg/daemon/daemon.go
```diff
@@ -422,6 +427,9 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 		}
 		//TODO HARDWARE PLUGIN for e810
 		if pProcess == ts2phcProcessName { //& if the x plugin is enabled
+			if output.gnss_serial_port == "" {
+				output.gnss_serial_port = GPSPIPE_SERIALPORT
```
jzding, line 431: shouldn't this be
```
output.gnss_serial_port = GPSDDefaultGNSSSerialPort
```

PR #172, opened 2023-08-02
OCPBUGS-16220: Replace Ublox with direct call to default gpsd  for monitoring and  reduce CPU usage 

aneeshkp:
This PR uses the direct call to gpsd for monitoring the GNSS state.
Current monitoring doesn't have GNSS offset. 
Once it is available then the offset will be reported accurately.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.19
 require github.com/openshift/ptp-operator v0.0.0-20230207052655-ede9197d99ca
 
 require (
+	github.com/aneeshkp/go-gpsd v0.0.0-20230804165302-4a2d069c566a
 	github.com/facebook/time v0.0.0-20230529151911-512b3b30ab23
 	github.com/golang/glog v1.0.0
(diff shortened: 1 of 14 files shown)


pkg/gpsd/gpsd.go
```diff
@@ -0,0 +1,317 @@
+package gpsd
```
josephdrichard, line 1: This is an external package?  Can this be vendored in instead?

aneeshkp: no i have modified to add  TOFF and other report in my updated PR .yet t push it

aneeshkp: and they are outdated. the struct values are in nano sec  not micro sec

aneeshkp: vendored via https://github.com/aneeshkp/go-gpsd

josephdrichard (approved):
/lgtm

PR #176, opened 2023-08-04
Clean up Dockerfiles to match upstream and downstream configuration

nishant-parekh:
* Delete Dockerfile.ocp
* Delete Dockerfile.rhel7


jzding (approved):
/lgtm

PR #180, opened 2023-08-08
OCPBUGS-17426: discard invalid dpll notifications

vitus133:
E810 DPLL driver may send incomplete DPLL notifications where, for example, LockStatus is omitted. 
When they are handled, the daemon reports freerun mode, while in fact the system is in holdover.
Invalid notifications should be discarded.
/cc @aneeshkp @jzding  @josephdrichard


aneeshkp (approved):
/lgtm

jzding (approved):
/lgtm

PR #181, opened 2023-08-08
OCPBUGS-17454: parse DPLL replies by type

vitus133:
This commit fixes a mistake in netlink dpll notification parsing. The parsing should be done by type and not by id
/cc @aneeshkp @jzding  @josephdrichard


aneeshkp (approved):
/lgtm

PR #185, opened 2023-08-11
use non-interactive pmc

jzding:
Use direct pmc commands instead of interactive mode to be more efficient.
Update ClockAccuracy and OffsetScaledLogVariance according to the standards when ClockClass is not changed.

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -356,42 +356,33 @@ func (e *EventHandler) updateCLockClass(cfgName string, ptpState PTPState, clock
 	case PTP_LOCKED:
 		switch clockType {
 		case GM:
-			// update only when ClockClass is changed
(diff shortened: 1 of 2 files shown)


pkg/daemon/daemon.go
```diff
@@ -314,8 +314,9 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 	dn.pluginManager.OnPTPConfigChange(nodeProfile)
 
 	ptp_processes := []string{
-		ts2phcProcessName,
+		// make sure ptp4l is running first to get pmc ready
```
josephdrichard, line 317: Will this cause issues with ptp4l propagating bad times before ts2phc sets the time to the correct time in a GM config?

PR #186, opened 2023-08-14
fix dpll termination

vitus133:
this commit fixes an error occurring when ptp profile is removed: panic: semaphore: released more than held


aneeshkp (approved):
/lgtm

PR #187, opened 2023-08-14
Fixed crashloop and clock class thread hijacking by using channel 

aneeshkp:
clockclass run via go rountine

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -34,6 +34,18 @@ const (
 // ClockType ...
 type ClockType string
 
+// ClockClassRequest ...
+type ClockClassRequest struct {
+	class     int8
+	cfgName   string
+	gmState   PTPState
+	clockType ClockType
+}
+
+var (
+	clockClassRequestCh = make(chan ClockClassRequest, 5)
+)
+
 const (
 	// GM ..
 	GM ClockType = "GM"
@@ -76,9 +88,6 @@ const (
 	PTP_UNKNOWN PTPState = "-1"
 )
 
-// CLOCK_CLASS_CHANGE ...
-const CLOCK_CLASS_CHANGE = "CLOCK_CLASS_CHANGE"
-
 const connectionRetryInterval = 1 * time.Second
 
 // EventHandler ... event handler to process events
@@ -92,6 +101,8 @@ type EventHandler struct {
 	data           map[string][]Data
 	offsetMetric   *prometheus.GaugeVec
 	clockMetric    *prometheus.GaugeVec
+	clockClass     fbprotocol.ClockClass
+	lastGmState    PTPState
 }
 
 // EventChannel .. event channel to subscriber to events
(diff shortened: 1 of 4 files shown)


pkg/event/event.go
```diff
@@ -170,12 +182,38 @@ connect:
...
 		}
 	}
+
+	if redialClockClass {
+		go func() {
+			for {
+				select {
+				case clk := <-clockClassRequestCh:
+					classErr, _ := e.updateCLockClass(&c, clk.cfgName, clk.gmState, clk.clockType)
+					if err != nil {
```
jzding, line 204: if classErr != nil

pkg/event/event.go
```diff
@@ -265,22 +303,17 @@ connect:
...
-				if err == nil {
-					// in case an error do not update lastGmState so that updateCLockClass can be tried again
-					lastGmState = gmState
-					logOut = append(logOut, fmt.Sprintf("%s[%d]:[%s] CLOCK_CLASS_CHANGE %d\n", PTP4l, time.Now().Unix(), event.CfgName, clockClass))
-				} else {
-					glog.Errorf("failed to updateClockClass: %v", err)
-				}
+				clockClassRequestCh <- ClockClassRequest{cfgName: event.CfgName, gmState: gmState, clockType: event.ClockType}
+				// in case an error do not update lastGmState so that updateCLockClass can be tried again
+				lastGmState = gmState
```
jzding, line 308: The lastGmState should only be updated when updateClockClass() was successful, otherwise if updateClockClass() failed the function will not be triggered again.

pkg/event/event.go
```diff
@@ -259,28 +305,22 @@ connect:
...
-				} else {
-					glog.Errorf("failed to updateClockClass: %v", err)
-				}
+
+			if lastGmState != gmState && (e.clockClass == protocol.ClockClassUninitialized || e.clockClassUpdateError == nil) {
+				// update clock class
+				glog.Infof("updating clock class for last clock class %#v and gmsState %s ", e.clockClass, gmState)
+				clockClassRequestCh <- ClockClassRequest{cfgName: event.CfgName, gmState: gmState, clockType: event.ClockType}
+				// in case an error do not update lastGmState so that updateCLockClass can be tried again
+				lastGmState = gmState
```
jzding, line 314: should check e.clockClassUpdateError  here before update lastGmState

pkg/event/event.go
```diff
@@ -259,28 +305,22 @@ connect:
...
-				err, clockClass := e.updateCLockClass(event.CfgName, gmState, event.ClockType)
-				if err == nil {
-					// in case an error do not update lastGmState so that updateCLockClass can be tried again
-					lastGmState = gmState
-					logOut = append(logOut, fmt.Sprintf("%s[%d]:[%s] CLOCK_CLASS_CHANGE %d\n", PTP4l, time.Now().Unix(), event.CfgName, clockClass))
-				} else {
-					glog.Errorf("failed to updateClockClass: %v", err)
-				}
+
+			if lastGmState != gmState && (e.clockClass == protocol.ClockClassUninitialized || e.clockClassUpdateError == nil) {
```
jzding, line 309: should not check clockClassUpdateError here. If clockclassupdate failed before we want to try it agian.

`if lastGmState != gmState || e.clockClass == protocol.ClockClassUninitialized`

aneeshkp: fixed

pkg/event/event.go
aneeshkp: fixed

jzding (approved):
/lgtm

PR #195, opened 2023-08-16
fix holdover logic

vitus133:
this commit fixes several issues in holdover timer, units and logic

//cc @aneeshkp @jzding @josephdrichard @nishant-parekh 

Tested and looks good. One thing that still doesn't work - when in holdover, clock class in metrics is 248.
In locked and freerun clock class was correct


aneeshkp (approved):
/lgtm

PR #198, opened 2023-08-22
OCPBUGS-17971: DPLL netlink API should retry sending events to event handler when send fails

aneeshkp:
Fix race conditions between monitoring process and event handler.- Event handler need to wait for side car socket to be available before process can be initiated. 
Now dpll  and gpsd Monitoring are called via event handler.
The dpll monitoring and gpsd monitoring calls will be registered with event-handler .
Once event-handler is available to serve, the Event-handler will run monitors

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -40,7 +40,7 @@ func GetKubeConfig() (*rest.Config, error) {
 		return configFromFlags(kubeConfig)
 	}
 
-	return nil, fmt.Errorf("Could not locate a kubeconfig")
+	return nil, fmt.Errorf("could not locate a kubeconfig")
 }
 
 type ProcessConfig struct {
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -59,7 +59,7 @@ type ptpProcess struct {
 	stopped           bool
 	logFilterRegex    string
 	cmd               *exec.Cmd
-	depProcess        []process // this could gpsd and other process which needs to be stopped if the parent process is stopped
+	depProcess        []process // these are list of dependent process which needs to be started/stopped if the parent process is starts/stops
 	nodeProfile       *ptpv1.PtpProfile
 	parentClockClass  float64
 	pmcCheck          bool
@@ -129,7 +129,7 @@ func New(
 	}
 	InitializeOffsetMaps()
 	pluginManager := registerPlugins(plugins)
-	eventChannel := make(chan event.EventChannel, 10)
+	eventChannel := make(chan event.EventChannel, 50)
 	return &Daemon{
 		nodeName:             nodeName,
 		namespace:            namespace,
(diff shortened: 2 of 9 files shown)


pkg/daemon/daemon.go
```diff
@@ -129,7 +129,7 @@ func New(
 	}
 	InitializeOffsetMaps()
 	pluginManager := registerPlugins(plugins)
-	eventChannel := make(chan event.EventChannel, 10)
+	eventChannel := make(chan event.EventChannel, 50)
```
josephdrichard, line 132: Why the change here?

pkg/daemon/gpsd.go
```diff
@@ -22,7 +22,7 @@ const (
 	GNSSMONITOR_INTERVAL = 5 * time.Second
 )
 
-type gpsd struct {
+type GPSD struct {
```
josephdrichard, line 25: Is there a functional reason this change is necessary?

pkg/config/config.go
```diff
@@ -40,7 +40,7 @@ func GetKubeConfig() (*rest.Config, error) {
 		return configFromFlags(kubeConfig)
 	}
 
-	return nil, fmt.Errorf("Could not locate a kubeconfig")
```
josephdrichard, line 43: remove?

pkg/daemon/daemon.go
```diff
@@ -59,7 +59,7 @@ type ptpProcess struct {
 	stopped           bool
 	logFilterRegex    string
 	cmd               *exec.Cmd
-	depProcess        []process // this could gpsd and other process which needs to be stopped if the parent process is stopped
+	depProcess        []process // this could GPSD and other process which needs to be stopped if the parent process is stopped
```
josephdrichard, line 62: If you're changing the comment, can you also fix the wording?  It's not really clear as is.

pkg/daemon/gpsd.go
```diff
@@ -187,20 +220,25 @@ retry:
...
+			CfgName:     g.processConfig.ConfigName,
 			IFace:       g.gmInterface,
 			Values: map[event.ValueType]int64{
 				event.GPS_STATUS: int64(tpv.Mode),
 				event.OFFSET:     g.offset,
 			},
-			ClockType:  processCfg.ClockType,
-			Time:       time.Now().Unix(),
+			ClockType:  g.processConfig.ClockType,
+			Time:       time.Now().UnixMilli(),
```
josephdrichard, line 234: Why is this UnixMilli and above is UnixNano?

pkg/daemon/daemon.go
aneeshkp: Just to catch any failures , Increased the buffere size of channel so , wont block until 50 events if happens to be in the queue

pkg/daemon/gpsd.go
aneeshkp: i had to export the gpsd struct so pubsub framework can have access to it.
// Subscriber ... event subscriber
type Subscriber struct {
	source     event.EventSource
	gpsd       *GPSD
	monitoring bool
	id         string
}

pkg/config/config.go
aneeshkp: No , the next line is not showing up ? I aded return nil, fmt.Errorf("could not locate a kubeconfig")
+ return nil, fmt.Errorf("could not locate a kubeconfig")

pkg/daemon/daemon.go
aneeshkp: done

pkg/daemon/gpsd.go
aneeshkp: shoud be all UnixMili , will fix if any missing

PR #203, opened 2023-08-29
OCPBUGS-18270: Add readme for e810 plugin

josephdrichard:


addons/intel/README.md
```diff
@@ -0,0 +1,109 @@
...
+    ts2phcOpts: " "
+    ts2phcConf: |
+      [nmea]
+      ts2phc.master 1
+      [global]
+      use_syslog  0
+      verbose 1
+      logging_level 7
+      ts2phc.pulsewidth 100000000
+      #GNSS module s /dev/ttyGNSS* -al use _0
```
aneeshkp, line 97: we can remove this line

addons/intel/README.md
```diff
@@ -0,0 +1,109 @@
...
+  namespace: openshift-ptp
+spec:
+  profile:
+  - name: "profile1"
+    ...
+    ...
+    ......   
+    plugins:
+      e810:
+        enableDefaultConfig: false
```
aneeshkp, line 29: can we make this as by default  false? at this point this is usable only for non WPC or if you want to use ts2phc only , correct ?

josephdrichard:
> do we have plugin setting for localMaxOffset and LocalHoldOver values ?

updated

addons/intel/README.md
josephdrichard: already by default is false

PR #204, opened 2023-08-29
OCPBUGS-17526: clock class is not set to 140 when DPLL is lost source and out of spec - from holdover

aneeshkp:
Defining HOLDOVER/LOCKED and FREERUN state from DPLL 
With Lost Source, IN Spec, and Out of Spec state transitions.
When GNSS is Lost, if the DPLL is in HOLDover and inSpec then do not set it to FREERUN, Instead, GM should be in Holdover state

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -36,15 +36,16 @@ type ClockType string
 
 // ClockClassRequest ...
 type ClockClassRequest struct {
-	class     int8
-	cfgName   string
-	gmState   PTPState
-	clockType ClockType
+	cfgName    string
+	gmState    PTPState
+	clockType  ClockType
+	clockClass fbprotocol.ClockClass
 }
 
 var (
-	// increased buffer due to slow reading
-	clockClassRequestCh = make(chan ClockClassRequest, 25)
+	//  make sure only one clock class update is tried if it fails next  try will pass
+	// this will also stop flooding
+	clockClassRequestCh = make(chan ClockClassRequest, 1)
 
 	PMCGMGetter = func(cfgName string) (protocol.GrandmasterSettings, error) {
 		cfgName = strings.Replace(cfgName, TS2PHCProcessName, PTP4lProcessName, 1)
@@ -107,6 +108,13 @@ const (
 
 const connectionRetryInterval = 1 * time.Second
 
+type grandMasterSyncState struct {
+	state      PTPState
+	clockClass fbprotocol.ClockClass
+	sourceLost bool
+	gmLog      string
+}
+
 // EventHandler ... event handler to process events
 type EventHandler struct {
 	sync.Mutex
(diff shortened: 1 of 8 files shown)


pkg/event/event.go
```diff
@@ -267,6 +298,10 @@ connect:
 
 	glog.Info("starting grandmaster state monitoring...")
 	lastGmState := PTP_NOTSET
+	select {
+	case readyCh <- true:
```
jzding, line 302: what does this do?

pkg/event/event.go
```diff
@@ -373,13 +418,15 @@ connect:
...
 
 			// update clock class
-			go func() {
-				select {
-				case clockClassRequestCh <- ClockClassRequest{cfgName: event.CfgName, gmState: gmState, clockType: event.ClockType}:
-				default:
-					glog.Error("clock class request not read")
-				}
-			}()
+			if gmState != lastGmState {
```
jzding, line 421: this will always be false due to  line 407 set lastGmState = gmState

pkg/event/event.go
```diff
@@ -452,9 +512,13 @@ func (e *EventHandler) updateCLockClass(cfgName string, ptpState PTPState, clock
...
-			// update only when ClockClass is changed
-			if g.ClockQuality.ClockClass != fbprotocol.ClockClass7 {
+			if e.isSourceLost(cfgName) && !e.outOfSpec && g.ClockQuality.ClockClass != fbprotocol.ClockClass7 {
+				g.ClockQuality.ClockClass = fbprotocol.ClockClass7
+				g.ClockQuality.ClockAccuracy = fbprotocol.ClockAccuracyUnknown
+				// T-REC-G.8275.1-202211-I section 6.3.5
+				g.ClockQuality.OffsetScaledLogVariance = 0xffff
+				err = gmSetterFn(cfgName, g)
+			} else if e.isSourceLost(cfgName) && e.outOfSpec && g.ClockQuality.ClockClass != protocol.ClockClassOutOfSpec {
 				g.ClockQuality.ClockClass = fbprotocol.ClockClass7
```
jzding, line 522: should be 
`g.ClockQuality.ClockClass = protocol.ClockClassOutOfSpec`

pkg/event/event.go
```diff
@@ -420,10 +555,10 @@ func (e *EventHandler) updateCLockClass(cfgName string, ptpState PTPState, clock
 		return err, clockClass
 	}
 	glog.Infof("current GRANDMASTER_SETTINGS_NP:\n%s", g.String())
-	switch ptpState {
-	case PTP_LOCKED:
-		switch clockType {
-		case GM:
+	switch clockType {
+	case GM:
```
josephdrichard, line 559: Can you add comments here what each clock class is?

pkg/event/event.go
aneeshkp: review again pls

pkg/event/event.go
aneeshkp: review again pls

pkg/event/event.go
aneeshkp: added

jzding (approved):
/lgtm

PR #209, opened 2023-09-12
OCPBUGS-18903: T-GM-STATUS not logged every sec  when dpll and gnss events were not reported  

aneeshkp:
fix GM Status to print  every sec for GM configuration


jzding (approved):
/lgtm

PR #224, opened 2023-10-10
OCPBUGS-20183: PTP4L output in T-GM 

aneeshkp:
set time property to true for clock class 6 and 7


pkg/event/event_test.go
```diff
@@ -35,8 +35,8 @@ func monkeyPatch() {
 				CurrentUtcOffsetValid: false,
```
jnunyez, line 35: Is this the equivalent to UTC_REASONABLE?

pkg/event/event_test.go
```diff
@@ -35,8 +35,8 @@ func monkeyPatch() {
...
 				Leap59:                false,
 				Leap61:                false,
-				TimeTraceable:         false,
-				FrequencyTraceable:    false,
+				TimeTraceable:         true,
+				FrequencyTraceable:    true,
 				PtpTimescale:          false,
 				TimeSource:            0,
```
jnunyez, line 41: Is this the equivalent to TimeSource in ptp4l packets? Should 0x20 (GPS) for T-GM when GNSS connected.

aneeshkp: updated to GNSS when connected . and set to unknown when GNSS is lost

pkg/event/event_test.go
aneeshkp: set CurrentUtcOffsetValid  to true for 6 and 7

jzding (approved):
/lgtm

PR #230, opened 2023-10-12
OCPBUGS-20514: PTP HW Clock Class after disabling NMEA sentences set to FREERUN

aneeshkp:
When NMEA string is lost but GPS is in sync ,set ts2phc to free run

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -29,7 +29,16 @@ const (
 	//Status           ValueType = "status"
 	PHASE_STATUS     ValueType = "phase_status"
 	FREQUENCY_STATUS ValueType = "frequency_status"
+	NMEA_STATUS      ValueType = "nmea_status"
 )
+
+var valueTypeHelpTxt = map[ValueType]string{
+	OFFSET:           "0 = FREERUN, 1 = LOCKED, 2 = HOLDOVER",
+	GPS_STATUS:       "0=NOFIX, 1=Dead Reckoning Only, 2=2D-FIX, 3=3D-FIX, 4=GPS+dead reckoning fix, 5=Time only fix",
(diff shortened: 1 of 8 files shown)


pkg/event/event.go
```diff
@@ -664,22 +677,41 @@ func (e *EventHandler) UpdateClockStateMetrics(state PTPState, process, iFace st
...
+					r := []rune(iface)
+					iface = string(r[:len(r)-1]) + "x"
+				}
 				for dataType, value := range processData { // update process with metrics
+					var dataValue float64
+					switch vType := value.(type) {
+					case int64:
+						dataValue = float64(vType)
+					case float64:
+						dataValue = vType
```
jzding, line 695: should be `dataValue = value`

pkg/event/event.go
```diff
@@ -756,7 +788,16 @@ func (e *EventHandler) updateData(event EventChannel) Data {
 	//if event.WriteToLog {
 	logData := make([]string, 0, len(event.Values))
 	for k, v := range event.Values {
-		logData = append(logData, fmt.Sprintf("%s %d", k, v))
+		switch vType := v.(type) {
+		case int64, int, int32:
+			logData = append(logData, fmt.Sprintf("%s %d", k, vType))
```
jzding, line 793: `vType` shoud be `v` here. Same for line 795 and 797.

pkg/event/event.go
```diff
@@ -817,7 +858,16 @@ func (e *EventHandler) addData(event EventChannel) Data {
...
```
jzding, line 863: vType should be v here. Same for line 865 and 867.

jzding (approved):
/lgtm

PR #232, opened 2023-10-23
OCPBUGS-22278:  Dpll update in RHEL 9.2 will not generate any DPLL events

vitus133:
@aneeshkp this is an attempt for manual dpll structure update.
If it doesn't work, will need to try and regenerate the files.


aneeshkp:
/retitle OCPBUGS-22278:  Dpll update in RHEL 9.2 will not generate any DPLL events

PR #234, opened 2023-11-03
dual ts2phc /dpll support for multi  card T-GM

aneeshkp:
This  code adds  updates  from 1 card GM  to support for multi Card GM

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"net"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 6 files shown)


pkg/event/event.go
```diff
@@ -369,10 +365,11 @@ func (e *EventHandler) updateGMState(cfgName string) grandMasterSyncState {
 		clockClass: e.gmSyncState[cfgName].clockClass,
 		sourceLost: e.gmSyncState[cfgName].sourceLost,
 	}
+	fmt.Printf("%#v", rGrandMasterSyncState)
```
jzding, line 368: is this log temporary?

jzding (approved):
/lgtm

pkg/event/stats.go
```diff
@@ -24,3 +22,32 @@ type DataMetrics struct {
...
+	signalSource EventSource // GNSS PPS
+	sourceLost   bool
+}
+
+func (d *Data) updateState() {
+	state := PTP_UNKNOWN
+	for _, detail := range d.Details { // 2 ts2phc or 2 dpll etc
+		switch detail.State {
+		case PTP_FREERUN: // if its free ru and main state is not holdover then this is the state
+			state = detail.State
```
jzding, line 43: if main state is holdever  and this detail.State is the last one in the slice then the main state will be updated to freerun which is not correct.

aneeshkp: Added stats_test  unit test and fixed it

jzding (approved):
/lgtm

PR #238, opened 2023-11-16
Add DPLL event subscription for  multiple cards

aneeshkp:
This PR extends DPLL  event subscription to  multiple card configured under WPC config .


pkg/config/config.go
```diff
@@ -18,13 +19,51 @@ const (
...
+func (i *IFaces) GetEventSource(iface string) event.EventSource {
+	for _, ii := range *i {
+		if ii.Name == iface {
+			return ii.Source
+		}
+	}
+	return event.PTP4l
+}
+
+// GetEventSource ... get event source
```
jzding, line 55: fix comment

pkg/daemon/config.go
```diff
@@ -176,31 +176,26 @@ func (output *ptp4lConf) populatePtp4lConf(config *string) error {
...
 
-func getSource(is_ts2phc_master string) (source event.EventSource, err error) {
-	if master, err := strconv.ParseBool(strings.TrimSpace(is_ts2phc_master)); err == nil {
-		if master {
-			return event.GNSS, nil
-		} else {
-			return event.PPS, nil
+func getSource(is_ts2phc_master string) event.EventSource {
+	if ts2phcmaster, err := strconv.ParseBool(strings.TrimSpace(is_ts2phc_master)); err == nil {
+		if err != nil || ts2phcmaster {
```
jzding, line 181: `err != nil` won't be true since it is in the path of  `err == nil` from previous line

aneeshkp: done

pkg/config/config.go
aneeshkp: done

jzding (approved):
/lgtm

PR #244, opened 2023-12-01
Fix event handling

jzding:
Fixed the following issues:
When an event is received with an iface not found in the stored in-memory data, the whole Data is re-created before adding DataDetails for this event and DataDetails for other ifaces are wiped out. As a result, there is only one iface event data is stored in-memory at any given time and the GM State calculation is only based on this one iface instead of multiple ifaces.

Another issue is the event data for dpll is stored using iface alias, while event data for gnss is stored using iface name, resulting same interface having multiple entries in in-memory data.


aneeshkp (approved):
/lgtm

PR #252, opened 2023-12-19
OCPBUGS-25682: False GM process with unknown interface appears after node reboot

aneeshkp:
Ignore unknown interface for GM state and wait until interface is available


josephdrichard (approved):
/lgtm
/approved

PR #253, opened 2023-12-20
OCPBUGS-17422: switch to continuously polling NAV-CLOCK

josephdrichard:


josephdrichard:
/hold
Still need to test

addons/intel/e810.go
```diff
@@ -50,6 +50,14 @@ done
 echo "Disabled all SMA and U.FL Connections"
 `
 
+func getDefaultUblxCmds() []E810UblxCmds {
+	defaultUblxCommands := E810UblxCmds{
+		ReportOutput: false,
+		Args:         []string{"-p", "CFG-MSG,1,34,1"},
```
jzding, line 56: add comments:
Set the receiver to output the message NAV-CLOCK once every nav solution.

jzding:
The NAV-STATUS and NAV-CLOCK are pulling at the same cadence, why not also change NAV-STATUS to continuous pulling?

pkg/ublox/ublox.go
```diff
@@ -226,19 +250,70 @@ func match(stdout string, ubLoxRegex *regexp.Regexp) (string, error) {
...
 
-// GetNavOffset ... get gnss offset
 func (u *UBlox) GetNavOffset() (string, error) {
-	args := []string{"-t", "-p", "NAV-CLOCK", "-P", "29.20"}
 
-	output, err := exec.Command(UBXCommand, args...).Output()
-	if err != nil {
-		return "", fmt.Errorf("error executing ubxtool command: %s", err)
-	}
+	output, _ := u.navPoll.UbloxPoll(1000000000, "tAcc")
```
jzding, line 255: use `0` to wait forever.

josephdrichard: Support for that (and also for larger numbers) was broken when I tested.  This should wait about 30 years so would be sufficient.

pkg/daemon/gpsd.go
```diff
@@ -52,8 +51,7 @@ type GPSDSubscriber struct {
 
 // Monitor ...
 func (s GPSDSubscriber) Monitor() {
```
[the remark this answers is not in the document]
josephdrichard: Required to implement  Monitor.  When removed:
*GPSDSubscriber does not implement "github.com/openshift/linuxptp-daemon/pkg/event".Subscriber (missing method Monitor)

josephdrichard:
> The NAV-STATUS and NAV-CLOCK are pulling at the same cadence, why not also change NAV-STATUS to continuous pulling?

done

jzding (approved):

pkg/daemon/gpsd.go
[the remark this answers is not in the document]
aneeshkp: we need Monito, monitor is called from event handler

pkg/daemon/gpsd.go
```diff
@@ -72,7 +71,8 @@ func (s GPSDSubscriber) Notify(source event.EventSource, state event.PTPState) {
 // MonitorProcess ... Monitor GPSD process
 func (g *GPSD) MonitorProcess(p config.ProcessConfig) {
 	g.processConfig = p
-	//go g.MonitorGNSSEventsWithGPSD(p, "E810")
+	glog.Info("Starting GNSS Monitoring")
+	go g.MonitorGNSSEventsWithUblox(p, "E810")
```
aneeshkp, line 75: this should move to monitoring

pkg/ublox/ublox.go
```diff
@@ -226,17 +211,51 @@ func match(stdout string, ubLoxRegex *regexp.Regexp) (string, error) {
...
+			u.active = false
+		}
+		go u.UbloxPollPushThread()
 	}
+}
 
-	offset := extractOffset(string(output))
-	return offset, nil
+func (u *UBlox) UbloxPollPushThread() {
+	for {
```
aneeshkp, line 245: why two for loop ?

aneeshkp (approved):
/lgtm

PR #271, opened 2024-02-27
OCPBUGS-29683: add DPLL phase offset

vitus133:
Add DPLL pin notification parser and handle pin notifications. Distinguish between pin notifications and device notifications. Update DPLL phase offset if included for the pin belonging to the DPLL instance.
/cc @aneeshkp @josephdrichard @jzding  @nishant-parekh


pkg/dpll-netlink/dpll.go
```diff
@@ -141,7 +142,7 @@ func ParseDeviceReplies(msgs []genetlink.Message) ([]*DoDeviceGetReply, error) {
 			case DPLL_A_MODULE_NAME:
 				reply.ModuleName = ad.String()
 			case DPLL_A_MODE:
-				reply.Mode = ad.Uint32()
+				//reply.Mode = ad.Uint32()
```
vitus133, line 145: We don't seem to use this mode anywhere

pkg/dpll-netlink/dpll.go
```diff
@@ -154,6 +155,8 @@ func ParseDeviceReplies(msgs []genetlink.Message) ([]*DoDeviceGetReply, error) {
 				reply.ClockId = ad.Uint64()
 			case DPLL_A_TYPE:
 				reply.Type = ad.Uint32()
+			default:
+				log.Println("default", ad.Type(), len(ad.Bytes()), ad.Bytes())
```
vitus133, line 159: If in the future API will change and we will fail to parse it, we'll see it in the logs much clearer than "netlink: attribute 5 is not a uint32; length: 8" as it was before this commit

pkg/dpll/dpll.go
```diff
@@ -334,14 +335,23 @@ func (d *DpllConfig) nlUpdateState(replies []*nl.DoDeviceGetReply) bool {
 			glog.Infof("discarding on clock ID %v (%s): ", nl.GetDpllStatusHR(reply), d.iface)
 		}
 	}
+	for _, pin := range pins {
+		// TODO - figure out pin ID per clock ID in the device plugin?
+		// Meanwhile there are only the relevant pins that report phase offset
```
vitus133, line 340: We need to find a way to know which pin IDs are relevant for every clock ID.
I don't see an elegant way to do it, other than provide it in a configuration. This will make it harder and more error-prone to configure the daemon, but will make the the intention explicit. 
Any other thoughts?

pkg/dpll/dpll.go
```diff
@@ -432,7 +461,7 @@ func (d *DpllConfig) MonitorDpllNetlink() {
 				goto abort
 			}
 
-			if d.nlUpdateState(replies) {
+			if d.nlUpdateState(replies, []*nl.DoPinGetReply{}) {
```
vitus133, line 464: Maybe we need to dump pins and get a real phase offset as well here?

pkg/dpll/dpll.go
```diff
@@ -585,7 +614,6 @@ func (d *DpllConfig) stateDecision() {
 			d.sourceLost = true
 		}
 		d.state = event.PTP_FREERUN
-		d.phaseOffset = FaultyPhaseOffset
```
vitus133, line 588: I removed all the phase offset mocking. I think it's no longer needed

vitus133 (commented):
Added some thoughts as comments to the code. Let's discuss

pkg/dpll-netlink/dpll.go
aneeshkp: we should remove this case; in case  if it exists and not used, we fall to default case ?

pkg/dpll/dpll.go
vitus133: I will try to look into the ICE code to see how they attain pin IDs.

pkg/dpll-netlink/dpll-uapi.go
```diff
@@ -3,9 +3,14 @@
...
 package dpll_netlink
 
-import "fmt"
+import (
+	"encoding/json"
+	"fmt"
+)
 
 const DPLL_MCGRP_MONITOR = "monitor"
+const DPLL_PHASE_OFFSET_DIVIDER = 1000
```
aneeshkp, line 12: where is this used ?DPLL_PHASE_OFFSET_DIVIDER

vitus133: d.phaseOffset = int64(math.Round(float64(phaseOffset / nl.DPLL_PHASE_OFFSET_DIVIDER / 1000)))
this is from pkg/dpll/dpll.go

pkg/dpll-netlink/dpll-uapi.go
```diff
@@ -3,9 +3,14 @@
...
+const DPLL_TEMP_DIVIDER = 1000
```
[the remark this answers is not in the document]
vitus133: Not used. But if we want to report device temperature one day this can become handy. These dividers are defined in dpll.yaml

pkg/dpll/dpll.go
vitus133: d.phaseOffset is now initialized in pin notification handler before calling this function. The offset is not updated when device notification is handled (it is not conveyed by the notification). I think in general it should be passed, but transients are possible. For example, when we get devices, phase offset is not yet initialized. But it will be in less than a second.

aneeshkp (approved):
/lgtm

PR #275, opened 2024-03-06
Add ha profile reader for phc2sys options to support PTP HA

aneeshkp:
This pull request introduces functionality to support configuring phc2sys with High Availability (HA) capability. The phc2sysOpts will be configured with profile names that it depends on during fail over. For example:
  `phc2sysOpts: -r -u 0 -m -N 8 -R 16 -n 24 -a -ha-profile profile1, profile2`

The code will search for -ha-profile and replace the profile1 and profile2 with socket paths:

`-ha-profile profile1, profile2 and  convert those to 
-z /var/run/ptp4l.0.socket -z /var/run/ptp4l.1.socket`

It will ignore  conversion if user sets the actual path  in the opts config. example `-z /var/run/ptp4l.0.socket -z /var/run/ptp4l.1.socket`

This enhancement allows users to create a separate standalone phc2sysconfig to configure HA. In a dual NIC setup, users can specify the phc2sysOptions using the above two methods."


pkg/daemon/daemon.go
```diff
@@ -352,14 +371,21 @@ func printNodeProfile(nodeProfile *ptpv1.PtpProfile) {
...
+	var phcProcess *phc2SysProcess
 
-	ptp_processes := []string{
-		ts2phcProcessName,
-		ptp4lProcessName,
-		phcProcessName,
+	ptpProcesses := []string{
+		ts2phcProcessName, // there can be only one ts2phc process in the system
+		ptp4lProcessName,  // there could be more than one ptp4l in the system
+		phcProcessName,    // there can be only one phc2sys process in the system
```
josephdrichard, line 388: phc2sysProcessName

pkg/daemon/daemon.go
```diff
@@ -389,11 +415,23 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 			messageTag = fmt.Sprintf("[ptp4l.%d.config]", runID)
 		case phcProcessName:
 			configInput = nodeProfile.Phc2sysConf
-			configOpts = nodeProfile.Phc2sysOpts
-			socketPath = fmt.Sprintf("/var/run/ptp4l.%d.socket", runID)
+			configOpts = nodeProfile.Phc2sysOpts // if ha_profile string found in options replace it before startign the process
```
josephdrichard, line 418: Remove comment

pkg/daemon/daemon.go
aneeshkp: Thats old code. Fixed it

pkg/daemon/daemon.go

pkg/daemon/daemon.go
```diff
@@ -352,14 +371,21 @@ func printNodeProfile(nodeProfile *ptpv1.PtpProfile) {
...
+/*
+update: March 7th 2024
+To support PTP HA phc2sys profile is appended to the end
+since phc2sysOpts needs to collect profile information from applied
+ptpconfig profiles for ptp4l
+*/
 func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile) error {
 
 	dn.pluginManager.OnPTPConfigChange(nodeProfile)
+	var phcProcess *phc2SysProcess
```
josephdrichard, line 383: phc2sysProcess

pkg/daemon/daemon.go
```diff
@@ -428,7 +466,10 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 		for index, section := range output.sections {
 			if section.sectionName == "[global]" {
 				section.options["message_tag"] = messageTag
-				section.options["uds_address"] = socketPath
+				// if ha profile is mentioned then you don't need uds_address here
+				if _, ok := (*nodeProfile).PtpSettings[PTP_HA_IDENTIFIER]; !ok {
```
josephdrichard, line 470: Can you create a function like (*nodeprofile).getPhc2SysHAConfig() that will parse the settings and return a clean list of HA configs if there are any, or an empty list otherwise?  Would be cleaner than just parsing through the PtpSetttings, and trimming whitespace.  And the more complexity we move out of this function, the better for us

pkg/daemon/daemon.go
```diff
@@ -558,10 +599,39 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 			printNodeProfile(nodeProfile)
 			return fmt.Errorf("failed to write the configuration file named %s: %v", configPath, err)
 		}
-		printNodeProfile(nodeProfile)
-		dn.processManager.process = append(dn.processManager.process, &dprocess)
```
josephdrichard, line 562: We're always doing this right, either on line 606 or 633?  Why is it necessary to handle like this?

pkg/daemon/daemon.go
```diff
@@ -387,13 +413,25 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 			configFile = fmt.Sprintf("ptp4l.%d.config", runID)
```
josephdrichard, line 413: Can we just keep a mapping of ptpconfigName: socketPath, and then there's no iterating through again around line 613?

pkg/daemon/daemon.go
```diff
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"cmp"
 	"fmt"
+	"k8s.io/utils/pointer"
```
josephdrichard, line 7: Why do we need this?

pkg/daemon/daemon.go
```diff
@@ -272,6 +282,15 @@ func (dn *Daemon) applyNodePTPProfiles() error {
 	glog.Infof("updating NodePTPProfiles to:")
 	runID := 0
 	slices.SortFunc(dn.ptpUpdate.NodeProfiles, func(a, b ptpv1.PtpProfile) int {
+		aHasPhc2sysOpts := a.Phc2sysOpts != nil && *a.Phc2sysOpts != ""
+		bHasPhc2sysOpts := b.Phc2sysOpts != nil && *b.Phc2sysOpts != ""
+		//sorted in ascending order
+		// here having phc2sysOptions is considered a low number
+		if !aHasPhc2sysOpts && bHasPhc2sysOpts {
+			return -1 //  a>b return -1
```
jzding, line 290: This is not correct, per https://pkg.go.dev/slices#SortFunc: 
cmp(a, b) should return a negative number when a < b, a positive number when a > b and zero when a == b.

aneeshkp: It is custom comparison.

pkg/daemon/daemon.go
aneeshkp: I do not want to refactor existing code. Makes hard for any backport fix, We should do iterative update

pkg/daemon/daemon.go
aneeshkp: again following existing code format; why treat this special ?

pkg/daemon/daemon.go
aneeshkp: This change it to build phc2sys at the end of the list, so that phc2sys process starts at the end

jzding (approved):
/lgtm

PR #276, opened 2024-03-10
CNF-11612: Add phase offset pin filter doc

vitus133:
Document an example of configuring phase offset pin filter in the example config

/cc @aneeshkp @josephdrichard


aneeshkp (approved):
/lgtm

PR #279, opened 2024-03-13
OCPBUGS-30856: Fix DPLL offset on holdover and freerun

vitus133:
This fixes a bug introduced in #271 leading to dpll phase offset freezing during trasitions to holdover and freerun states, where DPLL netlink indications from the driver stop. It also fixes the problem of not in-spec report on DPLL that has PPS as a source
/cc @aneeshkp @josephdrichard @jzding

### DIFF
--- a/pkg/dpll/dpll.go
+++ b/pkg/dpll/dpll.go
@@ -35,7 +35,7 @@ const (
 	LocalHoldoverTimeoutStr   = "LocalHoldoverTimeout"
 	MaxInSpecOffsetStr        = "MaxInSpecOffset"
 	ClockIdStr                = "clockId"
-	FaultyPhaseOffset         = math.MaxInt64
+	FaultyPhaseOffset         = 99999999999
 )
 
 type dpllApiType string
@@ -637,6 +637,7 @@ func (d *DpllConfig) stateDecision() {
 			d.sourceLost = true
 		}
 		d.state = event.PTP_FREERUN
+		d.phaseOffset = FaultyPhaseOffset
 		glog.Infof("dpll is in FREERUN, state is FREERUN (%s)", d.iface)
 		d.sendDpllEvent()
 	case DPLL_LOCKED:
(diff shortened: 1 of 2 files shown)


pkg/dpll/dpll.go
```diff
@@ -649,20 +650,24 @@ func (d *DpllConfig) stateDecision() {
 		} else { // what happens if source is lost and DPLL is locked? goto holdover?
 			glog.Infof("dpll is locked, offset is out of range, state is FREERUN(%s)", d.iface)
 			d.state = event.PTP_FREERUN
+			d.phaseOffset = FaultyPhaseOffset
```
aneeshkp, line 653: we don't see offset here , because DPLL is in locked state and the netlink event should have updated the offset

pkg/dpll/dpll.go
```diff
@@ -649,20 +650,24 @@ func (d *DpllConfig) stateDecision() {
...
+				d.phaseOffset = FaultyPhaseOffset
 				d.sourceLost = true
 			} else if dpllStatus == DPLL_LOCKED_HO_ACQ && d.isOffsetInRange() {
 				d.state = event.PTP_LOCKED
 				d.sourceLost = false
+				d.inSpec = true
 			} else {
 				d.state = event.PTP_FREERUN
 				d.sourceLost = false
+				d.phaseOffset = FaultyPhaseOffset
```
aneeshkp, line 670: same here

pkg/dpll/dpll.go
```diff
@@ -649,20 +650,24 @@ func (d *DpllConfig) stateDecision() {
...
 		if d.hasPPSAsSource() {
 			if dpllStatus == DPLL_HOLDOVER {
 				d.state = event.PTP_FREERUN // pps when moved to HOLDOVER we declare freerun
+				d.phaseOffset = FaultyPhaseOffset
 				d.sourceLost = true
 			} else if dpllStatus == DPLL_LOCKED_HO_ACQ && d.isOffsetInRange() {
 				d.state = event.PTP_LOCKED
 				d.sourceLost = false
+				d.inSpec = true
 			} else {
```
aneeshkp, line 667: dpll Status ins in LOCKED_HO_ACQ here you do get offsets ( set offset to faulty in case where you don't get offsets )

PR #282, opened 2024-03-15
PTPHA: update message tag to actual config file 

aneeshkp:
update message tag to actual config file so HA profile can be read from the config


jzding (approved):
/lgtm

PR #288, opened 2024-03-27
OCPBUGS-31370: fix ts2phc clock id from offset output and replace it with interfacename 

aneeshkp:
Problem:
PTP logs in Linux kernel versions before 4.2 referenced interfaces by name when ts2phc reports  offsets .
   In Linux 4.2 and later, PTP logs use a "ptp clock ID" instead of interface names for referencing.

Challenges:

    Existing tools and scripts that rely on parsing interface names in PTP logs might break with the newer format.
    Parsing logs consistently across different Linux versions becomes difficult.

Solution (Proposed by the PR):
   This PR aims to fix these challenges by providing functionality to:
   Convert "ptp clock ID" to the corresponding interface name.  Adjust the location of the "offset" field in the logs based on its position (potentially different between versions).
  This allows parsing older and newer PTP logs in a unified manner, regardless of the format used.


pkg/daemon/gpsd.go
```diff
@@ -247,29 +242,36 @@ retry:
 				}
 
 				//glog.Infof("MonitorGNSSEventsWithUblox nStatus=%d ; nOffset=%d", nStatus, nOffset)
-
 				g.offset = nOffset
-				if nStatus >= 3 && g.isOffsetInRange() {
+				g.sourceLost = false
+				if nStatus >= 3 {
 					g.state = event.PTP_LOCKED
```
jzding, line 248: nit: this line is not needed.

pkg/daemon/metrics.go
```diff
@@ -388,24 +388,31 @@ func extractRegularMetrics(configName, processName, output string) (err error, i
 	// for multi card GNSS
 	// ts2phc.0.cfg  /dev/ptp0  master    offset          0 s2 freq      -0
 	// ts2phc.0.cfg  /def/ptp4  master    offset          0 s2 freq      -0
+	// ts2phc.0.cfg  /def/ptp4 offset          0 s2 freq      -0
```
jzding, line 391: nit: replace /def/ with /dev/

pkg/network/utils.go
```diff
@@ -105,39 +106,42 @@ func DiscoverPTPDevices() ([]string, error) {
...
-		glog.Infof("GetPhcId: could not grab NIC timestamp capability for %s: %v", iface, err)
-		return
+		return -1, fmt.Errorf("failed to run ethtool: %w", err)
 	}
 
-	phcId = parsePhcId(&out)
-	if phcId != "" {
-		return fmt.Sprintf("/dev/%s", phcId)
+	// Regex to extract clock index
+	re := regexp.MustCompile(`PTP Hardware Clock: (\d+)`)
```
jzding, line 135: use _ETHTOOL_PTP_HARDWARE_CLOCK or delete the definition of _ETHTOOL_PTP_HARDWARE_CLOCK if not using it.

pkg/daemon/metrics.go
```diff
@@ -388,24 +388,31 @@ func extractRegularMetrics(configName, processName, output string) (err error, i
...
-		// Remove the element at index 1 from fields.
-		masterOffsetIface.set(configName, fields[1])
-		slaveIface.set(configName, fields[1])
-		copy(fields[1:], fields[2:])
-		// ts2phc.0.cfg  master    offset          0 s2 freq      -0
-		fields = fields[:len(fields)-1] // Truncate slice.
-		delay = 0
+	// linux 4.2 doesnt have master so we need to add that
+	//    0                1      2              3
+	//ts2phc.0.config  /dev/ptp6 offset          0 s2 freq      +0 (in linux 4.2)
```
jzding, line 398: this can be added as a new unit test case in daemon_test.go

pkg/network/utils.go
aneeshkp: done

pkg/daemon/metrics.go
aneeshkp: done

pkg/daemon/gpsd.go
aneeshkp: done

jzding (approved):
/lgtm

PR #296, opened 2024-04-02
OCPBUGS-31602: interfaces for deleted ptpconfig profiles still appear in metrics- process are not getting terminated 

aneeshkp:
The process restart was failing because the gpsd service wasn't shutting down properly (ending gracefully). This caused the cmdStop function to stall because gpsd was essentially hijacking it.
This Pull Request Fixes:
    Ensures gpsd terminates gracefully before the process restart, preventing it from blocking the cmdStop function.

### DIFF
--- a/pkg/daemon/gpsd.go
+++ b/pkg/daemon/gpsd.go
@@ -126,7 +126,7 @@ func (g *GPSD) CmdStop() {
 		}
 	}
 	g.unRegisterSubscriber()
-	<-g.exitCh
+	<-g.exitCh // waiting for all child routines to exit; we could add timeout to avoid waiting
 	glog.Infof("Process %s terminated", g.name)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/daemon/gpsd.go
```diff
@@ -236,6 +241,9 @@ retry:
 						}
 						break
 					}
+					if g.Stopped() {
```
josephdrichard, line 244: Which for loop is this intended to break from?

pkg/daemon/gpsd.go
```diff
@@ -126,7 +126,7 @@ func (g *GPSD) CmdStop() {
 		}
 	}
 	g.unRegisterSubscriber()
-	<-g.exitCh
+	<-g.exitCh // waiting for all child routines to exit; we could add timeout to avoid waiting
```
josephdrichard, line 129: Is it possible that MonitorGNSSEventsWithUblox will consume from this channel and then this will hang forever?

pkg/daemon/gpsd.go
aneeshkp: this loop 
for {
					//UbloxPollInit only initializes if not running
					ublx.UbloxPollInit()
					output := ublx.UbloxPollPull()

pkg/daemon/gpsd.go
aneeshkp: yes possible ; when one function reads from a single channel consumed by two functions, the other function will indeed hang waiting if monitoring is not  waiting on this channel. I will modify

pkg/daemon/gpsd.go
aneeshkp: fixed

josephdrichard: Can you move it out of the entire inner for loop entirely?  Maybe just after it instead?  The inner loop should be nearly instantaneous, so it seems a waste performance-wise to be grabbing and releasing the mutex so many times with the call to Stopped().

josephdrichard (approved):
/lgtm
/approved

PR #297, opened 2024-04-03
[release-4.15] OCPBUGS-xx: interfaces for deleted ptpconfig profiles still appear in metrics- process are not getting terminated

aneeshkp:
The process restart was failing because the gpsd service wasn't shutting down properly (ending gracefully). This caused the cmdStop function to stall because gpsd was essentially hijacking it.
This Pull Request Fixes:
Ensures gpsd terminates gracefully before the process restart, preventing it from blocking the cmdStop function.


jzding (commented):
/lgtm

PR #280, opened 2024-03-13
Add HA metrics to PTP HA 

aneeshkp:
`HELP openshift_ptp_ha_profile_status 0 = INACTIVE 1 = ACTIVE 0 # 
TYPE openshift_ptp_ha_profile_status gauge
openshift_ptp_ha_profile_status{node="node1",process="phc2sys",profile="profile1"} 1
 openshift_ptp_ha_profile_status{node="node1",process="phc2sys",profile="profile2"} 0`

PR #277, opened 2024-03-12
refactor HA profile to simplify damon

aneeshkp:

PR #302, opened 2024-04-26
[release-4.15] OCPBUGS-32305: fix missing openshift_ptp_process_status metric initial install

jzding:
This update port back some fixes around metrics cleanup and refactor from the following commits, except HA changes.
https://github.com/openshift/linuxptp-daemon/pull/275
https://github.com/openshift/linuxptp-daemon/pull/277
https://github.com/openshift/linuxptp-daemon/pull/280

Use sort.Slice() instead of slices.SortFunc() since the latter is not available in golang 1.20.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -51,15 +52,28 @@ type ProcessManager struct {
 
 // NewProcessManager is used by unit tests
 func NewProcessManager() *ProcessManager {
-	process := &ptpProcess{}
-	process.ptpClockThreshold = &ptpv1.PtpClockThreshold{
+	processPTP := &ptpProcess{}
+	processPTP.ptpClockThreshold = &ptpv1.PtpClockThreshold{
 		HoldOverTimeout:    5,
 		MaxOffsetThreshold: 100,
 		MinOffsetThreshold: -100,
 	}
 	return &ProcessManager{
-		process: []*ptpProcess{process},
+		process: []*ptpProcess{processPTP},
 	}
+}
+
+// SetTestProfile ...
+func (p *ProcessManager) SetTestProfileProcess(name string, ifaces config.IFaces, socketPath,
+	ptp4lConfigPath string, nodeProfile ptpv1.PtpProfile) {
+	p.process = append(p.process, &ptpProcess{
+		name:            name,
+		ifaces:          ifaces,
+		ptp4lSocketPath: socketPath,
+		ptp4lConfigPath: ptp4lConfigPath,
+		execMutex:       sync.Mutex{},
+		nodeProfile:     nodeProfile,
+	})
 }
 
 // SetTestData is used by unit tests
@@ -234,6 +248,11 @@ func printWhenNotNil(p interface{}, description string) {
 	}
 }
 
+// SetProcessManager ...
+func (dn *Daemon) SetProcessManager(p *ProcessManager) {
+	dn.processManager = p
+}
+
 func (dn *Daemon) applyNodePTPProfiles() error {
 	glog.Infof("in applyNodePTPProfiles")
 	for _, p := range dn.processManager.process {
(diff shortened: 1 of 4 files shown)


pkg/daemon/config.go
```diff
@@ -17,12 +17,29 @@ import (
...
+type nProfiles []ptpv1.PtpProfile
+
+// Len is required by sort.Interface
+func (p nProfiles) Len() int {
+	return len(p)
+}
+
+// Less is required by sort.Interface
+func (p nProfiles) Less(i, j int) bool {
+	return *p[i].Name < *p[j].Name
```
josephdrichard, line 29: This is being dropped from the sorting in master?
```
		aHasPhc2sysOpts := a.Phc2sysOpts != nil && *a.Phc2sysOpts != ""
		bHasPhc2sysOpts := b.Phc2sysOpts != nil && *b.Phc2sysOpts != ""
		//sorted in ascending order
		// here having phc2sysOptions is considered a high number
		if !aHasPhc2sysOpts && bHasPhc2sysOpts {
			return -1 //  a<b return -1
		} else if aHasPhc2sysOpts && !bHasPhc2sysOpts {
			return 1 //  a>b return
		}
```

jzding: yes we are not bringing HA related changes to 4.15

aneeshkp: @josephdrichard  aHasPhc2sysOpts && bHasPhc2sysOpts is HA releated

pkg/daemon/config.go
```diff
@@ -90,8 +107,8 @@ func (l *LinuxPTPConfUpdate) UpdateConfig(nodeProfilesJson []byte) error {
 }
 
 // Try to load the multiple policy config
-func tryToLoadConfig(nodeProfilesJson []byte) ([]ptpv1.PtpProfile, bool) {
-	ptpConfig := []ptpv1.PtpProfile{}
+func tryToLoadConfig(nodeProfilesJson []byte) (nProfiles, bool) {
+	ptpConfig := nProfiles{}
```
aneeshkp, line 111: I think its better we use same 
type in 4.16 also

aneeshkp: why not use same sorting method in 4.16 too @josephdrichard . So that we don't have too much code variation

pkg/daemon/config.go
aneeshkp: @josephdrichard sorry it is not HA related , It is generic. You are right  we should include that logic . @jzding

josephdrichard (commented):
/lgtm

pkg/daemon/daemon.go
```diff
@@ -51,17 +52,30 @@ type ProcessManager struct {
 
 // NewProcessManager is used by unit tests
 func NewProcessManager() *ProcessManager {
-	process := &ptpProcess{}
-	process.ptpClockThreshold = &ptpv1.PtpClockThreshold{
+	processPTP := &ptpProcess{}
```
aneeshkp, line 55: is this required to change ?

pkg/daemon/daemon.go
```diff
@@ -51,17 +52,30 @@ type ProcessManager struct {
...
 		MinOffsetThreshold: -100,
 	}
 	return &ProcessManager{
-		process: []*ptpProcess{process},
+		process: []*ptpProcess{processPTP},
 	}
 }
 
+// SetTestProfile ...
+func (p *ProcessManager) SetTestProfileProcess(name string, ifaces config.IFaces, socketPath,
```
aneeshkp, line 67: can this we done by creating a test file with same package name ?

aneeshkp: worried that we are  creating more new changes in  older version , we should carry these tests from master

pkg/daemon/daemon.go
```diff
@@ -234,6 +248,11 @@ func printWhenNotNil(p interface{}, description string) {
 	}
 }
 
+// SetProcessManager ...
+func (dn *Daemon) SetProcessManager(p *ProcessManager) {
```
aneeshkp, line 252: same here new introduction of a function that is not in new version

aneeshkp (approved):

PR #304, opened 2024-05-10
OCPBUGS-33524: PTP Boundary HA configuration cannot handle spaced empty Phc2SysOpts and ptp4lOpts with

aneeshkp:
Trim space of phc2sys and ptp4l config options.
If ptpConfig Opts are set with spaced empty string , then the daemon considers as value and tries to run the process for empty  spaced string


jzding (approved):
/lgtm

PR #306, opened 2024-05-22
OCPBUGS-34198: Revert "OCPBUGS-33524: PTP Boundary HA configuration cannot handle spaced empty Phc2SysOpts and ptp4lOpts with"

aneeshkp:
Reverts openshift/linuxptp-daemon#304
GM configuration is not applying correctly when setting ts2phc options to " " empty spaced string


jzding (approved):
/lgtm

PR #309, opened 2024-06-03
OCPBUGS-34804: fix leap file expiration and update times

vitus133:
This fixes a bug causing leap file expiration date to be off by 70 years and update date to be outdated


aneeshkp (approved):
/lgtm

PR #312, opened 2024-06-07
OCPBUGS-35029: linuxptp daemon taking almost 2 core with leap file updates

aneeshkp:


pkg/daemon/daemon.go
```diff
@@ -497,6 +497,9 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 					output.gnss_serial_port = strings.TrimSpace(gnssSerialPort)
 					section.options["ts2phc.nmea_serialport"] = GPSPIPE_SERIALPORT
 				}
+				if _, ok := section.options["leapfile"]; ok || pProcess == ts2phcProcessName { // not required to check process if leapfile is always included
```
jzding, line 500: shouldn't this be if not OK? so 
if leapfile option is not set (!ok) by customer, set default, 
if set by customer (ok), not use default.

pkg/daemon/gpsd.go
```diff
@@ -251,13 +251,17 @@ retry:
...
+						var lines []string
 						for i := 0; i < timeLsResultLines; i++ {
 							line := ublx.UbloxPollPull()
 							lines = append(lines, line)
 						}
 						timeLs := ublox.ExtractLeapSec(lines)
-						g.leapManager.UbloxLsInd <- *timeLs
+						select {
+						case g.leapManager.UbloxLsInd <- *timeLs:
+						case <-time.After(100 * time.Millisecond):
```
josephdrichard, line 262: Can you break this out of the inner for loop entirely?  Ideally after the event code below.  Otherwise, the reported information from the NAV-CLOCK and NAV-STATUS could be a bit old.  Also only want to do it once, not multiple times if there have been multiple of them in the ublox output.  Code at line 268 is to make sure we only look at the last occurrence of this.

pkg/daemon/daemon.go
vitus133: If leapfile is included in the config or process is ts2phc - override leapfile.
But since it can be included only in ts2phc config, can we just do 
``` 
if  pProcess == ts2phcProcessName 
```

PR #314, opened 2024-06-15
SyncE implementation 

aneeshkp:
This pull request includes the implementation to handle the SyncE process configured via ptpconfig. Key features and improvements are:

    Implemented SyncE process handling.
    Integrated configuration management through ptpconfig.
    Added necessary tests to ensure functionality and stability.


addons/intel/README.md
```diff
@@ -111,6 +111,9 @@ spec:
       [ens2f0]
       ts2phc.extts_polarity rising
       ts2phc.extts_correction 0
+    syncEOpts: " "
+    syncEConf: |
+      [global]
```
josephdrichard, line 116: synce4lOpts

pkg/daemon/daemon.go
```diff
@@ -461,6 +464,17 @@ func (dn *Daemon) applyNodePtpProfile(runID int, nodeProfile *ptpv1.PtpProfile)
 			configFile = fmt.Sprintf("ts2phc.%d.config", runID)
 			configPath = fmt.Sprintf("/var/run/%s", configFile)
 			messageTag = fmt.Sprintf("[ts2phc.%d.config:{level}]", runID)
+		case syncEProcessName:
+			configOpts = nodeProfile.Synce4lOpts
+			if configOpts == nil {
```
josephdrichard, line 469: This means that if a ptpconfig doesn't enable synce manually, it will still run synce.  This is the case with ptp4l, in order to be backwards compatible with other ptpconfigs, but for synce, if we don't specify Synce4lOpts(and so it's nil), then we should just not run synce.

aneeshkp: Right , I will fix this. Thanks

addons/intel/README.md
aneeshkp: fixed

josephdrichard (commented):
/lgtm
/approved

josephdrichard (approved):
/lgtm
/approved

PR #325, opened 2024-06-27
OCPBUGS-36201: implement leap event handling

vitus133:
this commit adds leap event handling that includes following:
1. Determine that current time is within -12h...1m from the leap
2. Send PMC command (once) to indicate leap61 or leap 59 to the downstream architecture and
3. Handle ts2phc time ambiguity


jzding (approved):
/lgtm

PR #329, opened 2024-07-10
assign clock IDs to SyncE devices

vitus133:
This commit adds automatic assignment of clock IDs to syncE devices.

### DIFF
--- a/addons/intel/e810.go
+++ b/addons/intel/e810.go
@@ -1,6 +1,7 @@
 package intel
 
 import (
+	"crypto/rand"
 	"encoding/binary"
 	"encoding/json"
 	"fmt"
@@ -89,7 +90,7 @@ func OnPTPConfigChangeE810(data *interface{}, nodeProfile *ptpv1.PtpProfile) err
 				glog.Error("e810 failed to unmarshal opts: " + err.Error())
 			}
 			if e810Opts.EnableDefaultConfig {
-				stdout, err = exec.Command("/usr/bin/bash", "-c", EnableE810PTPConfig).Output()
+				stdout, _ = exec.Command("/usr/bin/bash", "-c", EnableE810PTPConfig).Output()
 				glog.Infof(string(stdout))
 			}
 			if (*nodeProfile).PtpSettings == nil {
@@ -97,7 +98,15 @@ func OnPTPConfigChangeE810(data *interface{}, nodeProfile *ptpv1.PtpProfile) err
 			}
 			for device, pins := range e810Opts.DevicePins {
 				dpllClockIdStr := fmt.Sprintf("%s[%s]", dpll.ClockIdStr, device)
-				(*nodeProfile).PtpSettings[dpllClockIdStr] = strconv.FormatUint(getClockIdE810(device), 10)
+				// for unit testing only, PtpSettings may include "unitTest" key. The value is
+				// the path where resulting configuration files will be written, instead of /var/run
+				if _, found := (*nodeProfile).PtpSettings["unitTest"]; found {
+					buf := make([]byte, 8)
+					rand.Read(buf)
+					(*nodeProfile).PtpSettings[dpllClockIdStr] = strconv.FormatUint(binary.LittleEndian.Uint64(buf), 10)
+				} else {
+					(*nodeProfile).PtpSettings[dpllClockIdStr] = strconv.FormatUint(getClockIdE810(device), 10)
+				}
 				for pin, value := range pins {
 					deviceDir := fmt.Sprintf("/sys/class/net/%s/device/ptp/", device)
 					phcs, err := os.ReadDir(deviceDir)
--- a/pkg/daemon/config.go
+++ b/pkg/daemon/config.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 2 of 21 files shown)


addons/intel/e810.go
```diff
@@ -89,15 +90,21 @@ func OnPTPConfigChangeE810(data *interface{}, nodeProfile *ptpv1.PtpProfile) err
...
+				stdout, _ = exec.Command("/usr/bin/bash", "-c", EnableE810PTPConfig).Output()
 				glog.Infof(string(stdout))
 			}
 			if (*nodeProfile).PtpSettings == nil {
 				(*nodeProfile).PtpSettings = make(map[string]string)
 			}
 			for device, pins := range e810Opts.DevicePins {
 				dpllClockIdStr := fmt.Sprintf("%s[%s]", dpll.ClockIdStr, device)
-				(*nodeProfile).PtpSettings[dpllClockIdStr] = strconv.FormatUint(getClockIdE810(device), 10)
+				if _, found := (*nodeProfile).PtpSettings["unitTest"]; found {
```
josephdrichard, line 101: This is only relevant when running unit tests?

addons/intel/e810.go
```diff
@@ -89,15 +90,21 @@ func OnPTPConfigChangeE810(data *interface{}, nodeProfile *ptpv1.PtpProfile) err
...
+					buf := make([]byte, 8)
+					rand.Read(buf)
+					(*nodeProfile).PtpSettings[dpllClockIdStr] = strconv.FormatUint(binary.LittleEndian.Uint64(buf), 10)
```
josephdrichard, line 104: This code is to assign randomly-generated uuid?  Can you move this to a generate_uuid function?

pkg/daemon/testdata/synce-profile.json
```diff
@@ -0,0 +1,120 @@
...
+	"ptp4lOpts": "-2 --summary_interval -4",
+	"ptpSchedulingPolicy": "SCHED_FIFO",
+	"ptpSchedulingPriority": 10,
+	"synce4lConf": "[global]\nlogging_level              7\nuse_syslog                 0\nverbose                    1\nmessage_tag                [synce4l]\n[<synce1>]\nnetwork_option             2\nextended_tlv               1\nrecover_time               60\n#clock_id 5799633565433967748\nmodule_name                ice\ndnu_prio 0XFF\n[ens7f0]\ntx_heartbeat_msec          1000\nrx_heartbeat_msec          10\nallowed_qls                0xE\nallowed_ext_qls            0x40\n[{GNSS}]\nboard_label                GNSS-1PPS\ninput_QL                   0x1\ninput_ext_QL               0x20\n",
```
josephdrichard, line 115: Is it possible to have this use linebreaks so it shows up as multiple lined in this file?  Would be cleaner if we need to do single line changes in the future.

pkg/daemon/config.go
```diff
@@ -184,6 +185,93 @@ func getSource(isTs2phcMaster string) event.EventSource {
...
+	re, _ := regexp.Compile(`[{}<>\[\] ]+`)
+
+	for _, section := range conf.sections {
+		if strings.HasPrefix(section.sectionName, "[<") {
+			deviceName = re.ReplaceAllString(section.sectionName, "")
+			ifaces = []string{}
+		} else if strings.HasPrefix(section.sectionName, "[{") {
+			binding := synceToIfaceRelation{
+				synceDevice: deviceName,
+				ifaceNames:  ifaces,
```
josephdrichard, line 224: How do we know that ifaces will already be set?

addons/intel/e810.go
vitus133: No, this is not a UUID. ClockID is a UINT64 number. UUID package either gives a string, or a 16-byte buffer, that I would need to cut and turn to UINT64 anyway. Not sure this will simplify the code. WDYT?

pkg/daemon/config.go
vitus133: According to the comment above (regarding the config file section order), interface sections follow the device section (prefix "[<") and are followed by the port section (prefixed by  "[{" ). At the moment we reach this line, all the interfaces related to this device should be already processed (line 229).

pkg/daemon/testdata/synce-profile.json
vitus133: Unfortunately JSON doesn't allow real line breaks in the string data... But I like the idea - will try with yaml

vitus133 (commented):
Thanks @josephdrichard for your comments, added some thoughts. Please tell me what you think

addons/intel/e810.go
josephdrichard: Can you include a comment in the plugin explaining that?

pkg/daemon/testdata/synce-profile.json
vitus133: done

addons/intel/e810.go
vitus133: Done

pkg/daemon/config.go
vitus133: Thanks, I found a bug and fixed it. First of all, the external source section is optional - it might not be there for non-GM settings. But it is a requirement in synce4l documentation that device is followed by the interfaces.
Added a couple of unit tests to verify nothing breaks if the config file is wrong, and a log that says that clock id is not found because the interfaces are missing

vitus133 (commented):
@josephdrichard thanks for the review, made some changes

addons/intel/e810.go
vitus133: Or did you mean in the README?

josephdrichard (approved):
/lgtm
/approved

PR #335, opened 2024-07-18
OCPBUGS-37042: prevent writing uninitialized TimePropertiesDS

vitus133:
This PR fixes a bug where the daemon wrote uninitialized values to the time properties dataset, which led to corrupting the ptp time scale, current UTC offset and other dataset properties. The fix is done by always initializing the time properties according to the current UTC offset and a predefined set of flags. Leap manager was changed to provide current UTC offset to external packages.

### DIFF
--- a/pkg/daemon/gpsd.go
+++ b/pkg/daemon/gpsd.go
@@ -44,7 +44,6 @@ type GPSD struct {
 	subscriber           *GPSDSubscriber
 	monitorCtx           context.Context
 	monitorCancel        context.CancelFunc
-	leapManager          *leap.LeapManager
 }
 
 // GPSDSubscriber ... event subscriber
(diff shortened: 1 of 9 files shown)


pkg/daemon/gpsd.go
```diff
@@ -280,7 +279,7 @@ retry:
 				g.sourceLost = false
 				if timeLs != nil {
 					select {
-					case g.leapManager.UbloxLsInd <- *timeLs:
+					case leap.LeapMgr.UbloxLsInd <- *timeLs:
```
josephdrichard, line 282: This should probably be after the event processing code (switch nStatus >= 3).  If this blocks for a second or so, then it could lead to delays writing to the event channel, which in my opinion is more important to have realtime than updating the leap file.

pkg/leap/leap-file_test.go
```diff
@@ -299,3 +299,65 @@ func Test_handleLeapIndication(t *testing.T) {
 	assert.Equal(t, "3928780800", lm.leapFile.LeapEvents[1].LeapTime)
 	assert.Equal(t, "4291747200", lm.leapFile.ExpirationTime)
 }
+
+func Test_New_Good(t *testing.T) {
```
josephdrichard, line 303: Do we need to do any cleanup after this test case?

vitus133: Good one, done

josephdrichard (approved):
/ltgm
/approved

vitus133:
cherrypick release-4.16

PR #336, opened 2024-07-22
OCPBUGS-35587: T-GM - ts2phc process restart doesn't update PTP syncState when the process recovered

aneeshkp:
The ts2phc process termination only triggers FREERUN and doesn't recover back. This PR fixes this issue by sending the process status to the GM Event Decision Engine, which will mark T-GM as FREERUN. The event framework will use the PROCESS_DOWN indication to produce accurate events.


jzding (approved):
/lgtm

PR #341, opened 2024-07-24
[release-4.16] OCPBUGS-37043: prevent writing uninitialized TimePropertiesDS

vitus133:
This PR fixes a bug where the daemon wrote uninitialized values to the time properties dataset, which led to corrupting the ptp time scale, current UTC offset and other dataset properties. The fix is done by always initializing the time properties according to the current UTC offset and a predefined set of flags. Leap manager was changed to provide current UTC offset to external packages.

This is a manual cherrypick of #335 

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -130,7 +130,6 @@ func main() {
 		&refreshNodePtpDevice,
 		closeProcessManager,
 		cp.pmcPollInterval,
-		lm,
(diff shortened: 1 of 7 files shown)


aneeshkp:
/retitle [release-4.16] OCPBUGS-37043: prevent writing uninitialized TimePropertiesDS

PR #342, opened 2024-07-29
CNF12625: fix Leap second PMC command time window

vitus133:
This commit fixes leap second pmc command time window. Before the fix, pmc command is sent in a time window from -12h to +60s relatively to the leap event. The correct time window is -12h to -60s


aneeshkp (approved):
/lgtm

PR #347, opened 2024-08-06
fix clockid adding to synce device

aneeshkp:
ifxed bug that was preventing adding of Clock id  to synce config


vitus133:
/lgtm
A good one!

PR #357, opened 2024-09-01
OCPBUGS-39322: [release-4.18]fix ts2phc leap window check

vitus133:
This fixes the leap window condition check when receiving ts2phc status messages. Any of them should not cause "nmea string lost" event during the leap second clock irregularity.

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -966,10 +966,11 @@ func (p *ptpProcess) processPTPMetrics(output string) {
 		logEntry := synce.ParseLog(output)
 		p.ProcessSynceEvents(logEntry)
 
-	} else if p.name == ts2phcProcessName && (strings.Contains(output, NMEASourceDisabledIndicator) ||
-		strings.Contains(output, InvalidMasterTimestampIndicator) ||
-		(strings.Contains(output, NMEASourceDisabledIndicator2) &&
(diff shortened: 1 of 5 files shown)


pkg/daemon/daemon.go
```diff
@@ -949,10 +949,11 @@ func (p *ptpProcess) processPTPMetrics(output string) {
...
 
-	} else if p.name == ts2phcProcessName && (strings.Contains(output, NMEASourceDisabledIndicator) ||
-		strings.Contains(output, InvalidMasterTimestampIndicator) ||
-		(strings.Contains(output, NMEASourceDisabledIndicator2) &&
-			(!leap.LeapMgr.IsLeapInWindow(time.Now().UTC(), -2*time.Second, time.Second)))) { //TODO identify which interface lost nmea or 1pps
+	} else if p.name == ts2phcProcessName &&
+		!leap.LeapMgr.IsLeapInWindow(time.Now().UTC(), -2*time.Second, time.Second) &&
+		(strings.Contains(output, NMEASourceDisabledIndicator) ||
+			strings.Contains(output, InvalidMasterTimestampIndicator) ||
+			strings.Contains(output, NMEASourceDisabledIndicator2)) { //TODO identify which interface lost nmea or 1pps
```
aneeshkp, line 956: isn't nmea  lost indicator always for GM interface ? GM interface will  give the nmea interface , until we support two GNSS

vitus133: That's right, only for GM interface. I only added a condition for declaring "nmea string lost", that we are not in the leap event right now

PR #358, opened 2024-09-01
OCPBUGS-39187: [release-4.17] fix ts2phc leap window check

vitus133:
This fixes the leap window condition check when receiving ts2phc status messages. Any of them should not cause "nmea string lost" event during the leap second clock irregularity.


aneeshkp (approved):
/lgtm

aneeshkp (approved):
/lgtm

PR #367, opened 2024-09-11
OCPBUGS-19838:PTP HW Clock Class after disabling NMEA sentences - Handle T-GM holdover 

aneeshkp:
Holdover Enabled when GNSS time source  is lost :
The operator will  configure  ts2phcOpts: "--ts2phc.holdover ( timer calculated from DPLL )  --servo_offset_threshold  99999999 --servo_num_offset_values 10" 
The holdover should timeout same time as DPLL , hence we calculate DPLL timer to configure ts2phc holdover timeout 

     --ts2phc.holdover : The holdover interval, specified in seconds. 
When the ToD information stops working  (e.g. GNSS receiver lost its fix), `ts2phc` is allowed for the specified interval to continue synchronizing the target clock as long as the servo is in the `SERVO_LOCKED_STABLE` state. 
The servo state needs be enabled by the  `servo_offset_threshold` option. The holdover is not supported with the `-a` option and when `ts2phc.extts_polarity` is set to `both`. The default is 0 (disabled).
    
  --servo_num_offset_values:  The number of offset values considered in order to transition from the `SERVO_LOCKED` to the `SERVO_LOCKED_STABLE` state. The transition occurs once the last `servo_num_offset_values` offsets are all below the `servo_offset_threshold` value. The default value is 10. PTP Operator will set it to 99999 a large number to prevent switching 

servo_offset_threshold : The offset threshold used in order to transition from the `SERVO_LOCKED` to the `SERVO_LOCKED_STABLE` state.  The transition occurs once the last `servo_num_offset_values` offsets are all below the threshold value. The default value of offset_threshold is 0 (disabled). But this need to be enabled for holdover to work

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.22-openshift-4.17 AS builder
+FROM registry.ci.openshift.org/ocp/builder:rhel-9-golang-1.22-openshift-4.18 AS builder
 WORKDIR /go/src/github.com/openshift/linuxptp-daemon
 COPY . .
 RUN make clean && make
 
-FROM registry.ci.openshift.org/ocp/4.17:base-rhel9
+FROM registry.ci.openshift.org/ocp/4.18:base-rhel9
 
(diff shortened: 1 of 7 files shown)


jzding (approved):
/lgtm

jzding (approved):
/lgtm

PR #368, opened 2024-09-16
OCPBUGS-41882: Disable Ublox GNGSA and GNGSV nmea messages

vitus133:
This commit disables SA and SV messages to reduce unnecessare computing load on ts2phc. The amount of these two messages can increase uncontrollably with the number of visible satellites, and they don't provide any useful information for the timing solution.


aneeshkp (approved):
/lgtm

PR #372, opened 2024-11-04
fix fips complaint issue

aneeshkp:
Fix fips complaint issue to enable fips support 
Go compliance shim [14] [openshift-4.18][linuxptp-daemon]: non-compliant: eliminated no_openssl


jzding (approved):
/lgtm

PR #374, opened 2024-11-20
Add missing LISTENING port states to the list of states reported in metrics

edcdavid:
This is to add support for the LISTENING port state in metrics. Using a new integer value 5 for backward compatibility (so as to not change previous mappings). 
The LISTENING state is reached when multiple follower ports are receiving PTP frames but only one is selected to become a SLAVE, in slave only mode. The other ports state should be LISTENING (whether they receive PTP frames are received or not).

### DIFF
--- a/pkg/daemon/metrics.go
+++ b/pkg/daemon/metrics.go
@@ -64,6 +64,7 @@ const (
 	MASTER
 	FAULTY
 	UNKNOWN
+	LISTENING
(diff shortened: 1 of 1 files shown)


aneeshkp:
you might have to do the same to cloud-event-proxy

aneeshkp (approved):
/lgtm

PR #376, opened 2024-12-03
OCPBUGS-45266: Reduce unwanted NMEA messages

vitus133:
This commit reduces the number of NMEA messages going to the log and through the `ts2phc` to the essential two message types: GNRMC and GNGGA.

### DIFF
--- a/addons/intel/e810.go
+++ b/addons/intel/e810.go
@@ -76,12 +76,33 @@ func getDefaultUblxCmds() []E810UblxCmds {
 		ReportOutput: false,
 		Args:         []string{"-p", "CFG-MSG,0xf0,0x03,0"},
 	}
+	// Ublx command to disable VTG messages
+	cfgMsgDisableVTG := E810UblxCmds{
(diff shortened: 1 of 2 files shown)


addons/intel/e810.go
```diff
@@ -76,12 +76,33 @@ func getDefaultUblxCmds() []E810UblxCmds {
 		ReportOutput: false,
 		Args:         []string{"-p", "CFG-MSG,0xf0,0x03,0"},
 	}
+	// Ublx command to disable VTG messages
+	cfgMsgDisableVTG := E810UblxCmds{
```
aneeshkp, line 80: why not make this changes in PTPconfig ? there is more flexibility  for changes that way than putting them in code ?

vitus133: I think this is not something users would like to tweak with. As opposed to NAV-TIMELS message, that users of other constellations than GPS might want to disable, these messages do nothing for T-GM but harm.

PR #383, opened 2024-12-18
OCPBUGS-46370: GM clock class transition to freerun

vitus133:
This covers the case where LocalHoldoverTimeout is smaller than the one defined by MaxInSpecOffset.


jzding (approved):
/lgtm

PR #404, opened 2025-02-20
OCPBUGS-51115: [4.16]Port 4.14 changes

vitus133:
This is a missing port from #379 
Remove clock class 140, add debug package and prints
/cc @aneeshkp @josephdrichard @jzding


nocturnalastro (approved):
/lgtm

PR #412, opened 2025-03-28
OCPBUGS-53371: Increasing Clock Class Reporting Frequency

aneeshkp:
For Boundary Clock, the GM clock class is read via the PMC command every few seconds and logged only when there is a change in CLOCK CLASS. This causes the sidecar to miss updates on the clock class value if it restarts or fails to read the first event. Increasing the frequency ensures that the latest clock class is always updated.


jzding (approved):
/lgtm

PR #428, opened 2025-04-21
OCPBUGS-55163: [release-4.16] keep in holdover until locked-holdover-acquired

vitus133:
This is a manual cherry-pick of #413
Simplify DPLL eventing logic by decoupling holdover and LHAQ states Remove "locked" DPLL state handling, keeping in Holdover until the LHAQ state reached.
Fix a corner case when holdover timer keeps running when DPLL goes to unlocked.
Fix a race condition in tests where event source was not fast enough to be set before the event processing

/assign @aneeshkp 
/cc @josephdrichard @jzding @nocturnalastro

### DIFF
--- a/pkg/dpll/dpll.go
+++ b/pkg/dpll/dpll.go
@@ -10,12 +10,14 @@ import (
 	"sync"
 	"time"
 
+	ptpv1 "github.com/openshift/ptp-operator/api/v1"
+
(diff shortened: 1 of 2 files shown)


vitus133:
/test e2e-aws
(infrastructure failure on the previous run)

PR #432, opened 2025-04-29
OCPBUGS-55466: [release-4.15] keep in holdover until locked-holdover-acquired

vitus133:
Manual cherrypick of #428 
Simplify DPLL eventing logic by decoupling holdover and LHAQ states Remove "locked" DPLL state handling, keeping in Holdover until the LHAQ state reached.
Fix a corner case when holdover timer keeps running when DPLL goes to unlocked.
Fix a race condition in tests where event source was not fast enough to be set before the event proces
/assign @aneeshkp


aneeshkp (approved):
/lgtm

PR #435, opened 2025-04-30
[release-4.16] OCPBUGS-55281: update event to fix T-GM regression failure

aneeshkp:
This PR restores a critical logic block that was incorrectly removed in a previous commit. The removed code was responsible for:

    missing  logDataValues from dataDetails.logData for log output
    if event.WriteToLog && logDataValues != "" {
	logOut = append(logOut, logDataValues)
}

### DIFF
--- a/pkg/event/event_test.go
+++ b/pkg/event/event_test.go
@@ -73,10 +73,10 @@ func TestEventHandler_ProcessEvents(t *testing.T) {
 			outOfSpec:        false,
 			iface:            "ens1f0",
 			values:           map[event.ValueType]interface{}{event.OFFSET: 0, event.PHASE_STATUS: 3, event.FREQUENCY_STATUS: 3, event.PPS_STATUS: 1},
-			wantGMState:      "GM[0]:[ts2phc.0.config] unknown T-GM-STATUS s0",
-			wantClockState:   "ptp4l[0]:[ts2phc.0.config] CLOCK_CLASS_CHANGE 248",
+			wantGMState:      "",
+			wantClockState:   "",
 			wantProcessState: "dpll[0]:[ts2phc.0.config] ens1f0 frequency_status 3 offset 0 phase_status 3 pps_status 1 s2",
-			desc:             "Initial state, gnss is not set, so expect GM to be FREERUN",
+			desc:             "Initial state, gnss is not set, gm won't have intrerface so no  t-gm or clock class ",
 		},
 		{
 			processName:      event.GNSS,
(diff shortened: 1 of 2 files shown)


pkg/event/event_test.go
```diff
@@ -307,23 +313,28 @@ func TestEventHandler_ProcessEvents(t *testing.T) {
...
-				if strings.HasPrefix(c, "ptp4l[") {
-					assert.Equal(t, test.wantClockState, rs, test.desc)
+				// Mark expected log as seen
+				for want := range expectedLogs {
+					if want == "" || c == want {
+						expectedLogs[want] = true
+					}
 				}
 			default:
+				time.Sleep(100 * time.Millisecond)
```
lack, line 327: I don't want to derail the PR, but I'd love to learn more about why we're sleeping here.

Every time I see a sleep I see a red flag; trauma from past jobs I've held.

pkg/event/event_test.go
```diff
@@ -73,10 +73,10 @@ func TestEventHandler_ProcessEvents(t *testing.T) {
 			outOfSpec:        false,
 			iface:            "ens1f0",
 			values:           map[event.ValueType]interface{}{event.OFFSET: 0, event.PHASE_STATUS: 3, event.FREQUENCY_STATUS: 3, event.PPS_STATUS: 1},
-			wantGMState:      "GM[0]:[ts2phc.0.config] unknown T-GM-STATUS s0",
-			wantClockState:   "ptp4l[0]:[ts2phc.0.config] CLOCK_CLASS_CHANGE 248",
+			wantGMState:      "",
+			wantClockState:   "",
 			wantProcessState: "dpll[0]:[ts2phc.0.config] ens1f0 frequency_status 3 offset 0 phase_status 3 pps_status 1 s2",
-			desc:             "Initial state, gnss is not set, so expect GM to be FREERUN",
+			desc:             "Initial state, gnss is not set, gm won't have intrerface so no  t-gm or clock class ",
```
lack, line 79: nit:
```suggestion
			desc:             "Initial state, gnss is not set, gm won't have interface so no t-gm or clock class ",
```

lack:
As someone new to the codebase, it would be useful to have a bit more in the commit log message - How does it address the T-GM regression failure?

pkg/event/event_test.go
nocturnalastro: Sleeping allows other go-routines to run which is especially important for the tests at they may run in environments for where the thread/process count is low :)

PR #436, opened 2025-05-05
OCPBUGS-55687: No event when  gpsd process restarted

aneeshkp:
The issue was that linuxptp logged severity-level errors to stderr, and the NMEA string timeout was being treated as an error. Our application wasn't capturing stderr, so it was missing those messages. I've now updated it to pipe both stderr and stdout to the same reader.
Now this was corrceted in 4.18 and above, but not found in 4.17 . Log severity was introduced in 4.16


josephdrichard (approved):
/lgtm

PR #475, opened 2025-08-04
OCPBUGS-59580: [release-4.17]Set clock accuracy per clock class in T-GM

vitus133:
/cc @aneeshkp @josephdrichard @nocturnalastro @jzding


aneeshkp (approved):
/lgtm

PR #552, opened 2026-02-24
OCPBUGS-80949,OKD-322: Update Dockerfile to work on CentOS/RHEL 10

jatinsu:
The OKD team is attempting to use CentOS 10 on 4.22. While attempting to build this, liinuxptp-daemon was failing to build due to gpsd-minimal and gpsd-minimal-clients not being a valid package. This is due to gpsd-minimal being renamed to gpsd. See [rhel 10 docs](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/10/html/considerations_in_adopting_rhel_10/infrastructure-services). This PR checks if the base OS is CentOS/RHEL 10, then installs the correct package based on the version.

### DIFF
--- a/addons/redhat/Dockerfile.ocp
+++ b/addons/redhat/Dockerfile.ocp
@@ -16,8 +16,9 @@ RUN yum -y update && yum -y update glibc && yum --setopt=skip_missing_names_on_i
   iproute \
   procps-ng \
   chrony \
-  gpsd-minimal \
-  gpsd-minimal-clients \
+  # gpsd-minimal and gpsd-minimal-clients were renamed to gpsd and gpsd-clients in RHEL 10
+  # try the old names first, fall back to new names if it fails
+  && (yum -y install gpsd-minimal gpsd-minimal-clients || yum -y install gpsd gpsd-clients) \
   && yum clean all
 
 # Create symlinks for executables to match references


nocturnalastro:
Sorry I was away on thursday, friday just catching up. I think if we move the install after the update I'm fine with the install method. Perhaps we could try something like 
```
yum update && \
yum install other dependencies && \
(yum install <rhel9 packages> || yum install <rhel10 packages>)
```
Would something like this be neater? 
Sorry for yet another solution. 
Just wondering if we can get away without sorting the names.

jatinsu:
@nocturnalastro no worries! Would something like this work?
```
RUN yum -y update && yum -y update glibc && yum --setopt=skip_missing_names_on_install=False -y \
  install \
  pciutils \
  linuxptp \
  ethtool \
  hwdata \
  synce4l \
  iproute \
  procps-ng \
  chrony \
  && (yum -y install gpsd-minimal gpsd-minimal-clients || yum -y install gpsd gpsd-clients) \
  && yum clean all
```

Seemed to work on rhel9 and CentOS 10.

image

jatinsu:
@nocturnalastro Just wanted to poke you in case this flew under the radar for you :)

nocturnalastro:
Thanks for the ping year sorry I'd missed to notification. 
This looks good :)

jatinsu:
@nocturnalastro Just commited, just waiting for a lgtm/approval again. I think this is the first time I've interacted with verifying PRs with OCPBUGS. What is the typical workflow after you give approval?

nocturnalastro:
@jatinsu Each time you push it clears the approval so I just need to add the member QE and re-do the approval.

nocturnalastro:
/retitle OCPBUGS-80949,[OKD-322](https://redhat.atlassian.net/browse/OKD-322): Update Dockerfile to work on CentOS/RHEL 10

PR #555, opened 2026-02-26
OCPBUGS-77473: Fix offset spike caused by not full window in holdover

nocturnalastro:
Instead use the caculated offset as it should be when in holdover

Co-Author: Vitaly Grinberg <vgrinber@redhat.com>


jzding (approved):
/lgtm

PR #570, opened 2026-03-20
[release-4.21] OCPBUGS-77871: Fix clock class metrics lost after cloud-event-proxy restart

jzding:
This is a backport from https://github.com/openshift/linuxptp-daemon/commit/e28b1904c114ea9df240a8b584be1ff6e7383a1d

When cloud-event-proxy crashes and restarts, clock_class metrics for most ptp4l configs disappear because:

1. EmitClockClassLogs() skipped configs where pmc.parentDS was nil, even though the clock class data was available in clkSyncState. Remove this unnecessary guard since EmitClockClass() already handles missing data gracefully.

2. emitClockClass() used utils.EmitClockClass() which writes directly to the socket without reconnect-and-retry logic. On broken pipe the data was silently lost. Switch to writeLogToSocket() which handles reconnection, consistent with EmitClockSyncLogs, EmitPortRoleLogs, and EmitProcessStatusLog.

3. Clean up now-unused code: utils.EmitClockClass, utils.IsBrokenPipe, signalBrokenPipe, brokenPipeCh, and their associated tests

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -206,12 +206,7 @@ func (p *ProcessManager) EmitClockClassLogs() {
 			for _, dp := range proc.depProcess {
 				if dp.Name() == PMCProcessName {
 					pmc := dp.(*PMCProcess)
-					if pmc.parentDS != nil {
-						// if parentDS is nil that means the clock class will
-						// be announced as soon as we get one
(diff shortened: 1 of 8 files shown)


nocturnalastro (approved):
/lgtm

PR #573, opened 2026-03-24
[release-4.21] OCPBUGS-77871: Fix race condition in writeLogToSocket dropping writes during reconnect

jzding:
Cherry-pick from https://github.com/k8snetworkplumbingwg/linuxptp-daemon/pull/184

Fixing two issues caused clock_class metrics to be missing after cloud-event-proxy crashed and recovered:

1. writeLogToSocket returned false when conn was nil (set by another
   goroutine handling a broken pipe) without attempting reconnection.
   Concurrent writers like EmitClockClassLogs silently lost data.
   Fix: call reconnectEventSocket when conn is nil, blocking until
   any in-progress reconnection completes.

2. clkSyncState was never populated with clock class values in T-BC/HA
   configurations. The clockClassRequestCh handler and UpdateClockClass
   set e.clockClass (EventHandler level) but not clkSyncState entries.
   EmitClockClass and the classTicker rely on clkSyncState for
   re-emission, so they had nothing to emit.
   Fix: store clock class in clkSyncState when received via
   clockClassRequestCh, via new storeClockClassLocked helper.

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -630,6 +630,17 @@ func (e *EventHandler) setClockClassLocked(clockClass fbprotocol.ClockClass, clo
 	e.clockAccuracy = clockAcc
 }
 
+// storeClockClassLocked stores the clock class and accuracy in clkSyncState
+// so that EmitClockClass and the classTicker can re-emit after a reconnect.
+// Caller must hold e.Lock().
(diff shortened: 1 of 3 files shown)


nocturnalastro (approved):
/lgtm

PR #576, opened 2026-03-26
OCPBUGS-80924: [4.21] Allow non-GNSS leap second sources

vitus133:
Cherry-pick
Add API to enable / disable leap second sources.
By default (and if leapSources is omitted) all satellite sources are enabled. To disable sources, specify them under the plugin gnss->leapSources section:
```
    e825:
      devices:
        - eno8703 
      gnss: 
        disabled: false 
        leapSources: 
          navic: false
```

### DIFF
--- a/addons/intel/common.go
+++ b/addons/intel/common.go
@@ -1,0 +1,62 @@
+// Package intel contains plugins for all supported Intel NICs
+package intel
+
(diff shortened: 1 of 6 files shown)


jzding (approved):
/lgtm

PR #585, opened 2026-04-09
[release-4.20] OCPBUGS-77881: Fix T-BC clock class not re-emitted after cloud-event-proxy restart

jzding:
## Problem

  After cloud-event-proxy process is killed and restarts in T-BC mode, clock class
  metrics (`openshift_ptp_clock_class`) are lost and never recovered. The 60-second
  classTicker that should periodically re-emit clock class events produces no output.

  ## Root Cause

  `announceClockClass` calls `setClockClassLocked` (which sets `e.clockClass`) and
  `emitClockClass` (which writes to the socket), but never `storeClockClassLocked`.
  After cloud-event-proxy restarts, `EmitClockClass` and the classTicker check
  `clkSyncState` and find it empty for ptp4l configs, so clock class is never
  re-emitted.

  ## Fix

  Add `storeClockClassLocked` call to `announceClockClass` so `clkSyncState` is
  populated on every clock class announcement. The classTicker's existing `SplitN`
  logic normalizes `ts2phc` config names to `ptp4l` for emission, so no changes
  are needed in `event_tbc.go` or `downstreamAnnounceIWF`.

  ## Testing

  - Verified on helix98 T-BC cluster: after killing cloud-event-proxy process,
    clock class metrics recover and the classTicker emits `CLOCK_CLASS_CHANGE`
    events every 60 seconds.
  - Unit tests:
    - `TestAnnounceClockClass_PopulatesClkSyncState` — verifies announceClockClass stores in clkSyncState and emits to socket

### DIFF
--- a/pkg/event/event.go
+++ b/pkg/event/event.go
@@ -615,6 +615,7 @@ func (e *EventHandler) AnnounceClockClass(clockClass fbprotocol.ClockClass, cloc
 func (e *EventHandler) announceClockClass(clockClass fbprotocol.ClockClass, clockAcc fbprotocol.ClockAccuracy, cfgName string) {
 	e.Lock()
 	e.setClockClassLocked(clockClass, clockAcc)
+	e.storeClockClassLocked(cfgName, clockClass, clockAcc)
 	e.Unlock()
 
 	e.emitClockClass(clockClass, cfgName)
--- a/pkg/event/event_socket_test.go
+++ b/pkg/event/event_socket_test.go
@@ -788,3 +788,46 @@ func TestStoreClockClassLocked_MakesEmitClockClassWork(t *testing.T) {
 
 	e.setConn(nil) // cleanup
 }
+
+// --- announceClockClass stores in clkSyncState ---
+
+func TestAnnounceClockClass_PopulatesClkSyncState(t *testing.T) {
+	socketPath := shortSocketPath(t)
+	listener, err := net.Listen("unix", socketPath)
+	assert.NoError(t, err)
+	defer listener.Close()
+
+	received := make(chan string, 10)
+	go acceptAndRead(listener, received)
+
+	e := newTestEventHandler(socketPath)
+	assert.True(t, e.reconnectEventSocket())
+
+	// clkSyncState should be empty initially
+	e.Lock()
+	_, ok := e.clkSyncState["ptp4l.1.config"]
+	e.Unlock()
+	assert.False(t, ok, "clkSyncState should be empty before announceClockClass")
+
+	// announceClockClass should populate clkSyncState AND emit to socket
+	e.announceClockClass(fbprotocol.ClockClass(6), fbprotocol.ClockAccuracy(0x21), "ptp4l.1.config")
+
+	// Verify clkSyncState was populated
+	e.Lock()
+	state, ok := e.clkSyncState["ptp4l.1.config"]
+	e.Unlock()
+	assert.True(t, ok, "clkSyncState should have ptp4l.1.config after announceClockClass")
+	assert.Equal(t, fbprotocol.ClockClass(6), state.clockClass)
+	assert.Equal(t, fbprotocol.ClockAccuracy(0x21), state.clockAccuracy)
+
+	// Verify it was written to socket
+	select {
+	case got := <-received:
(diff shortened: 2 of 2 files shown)


pkg/event/event.go
```diff
@@ -789,7 +796,7 @@ func (e *EventHandler) ProcessEvents() {
 					for clkCfgName, clockClass := range clkSnapshot {
 						parts := strings.SplitN(clkCfgName, ".", 2)
 						if len(parts) >= 2 {
-							clkCfgName = "ptp4l." + strings.Join(parts[1:], ".")
+							clkCfgName = "ptp4l." + parts[1]
```
nocturnalastro: Pretty sure this is wrong it should be `ptp4l.<runID>.config`

pkg/event/event_tbc.go
```diff
@@ -386,7 +386,7 @@ func (e *EventHandler) downstreamAnnounceIWF(cfgName string) {
 		StepsRemoved:        results.CurrentDS.StepsRemoved,
 	}
 	glog.Infof("%++v", es)
-	e.announceClockClass(gs.ClockQuality.ClockClass, gs.ClockQuality.ClockAccuracy, cfgName)
+	e.AnnounceClockClass(gs.ClockQuality.ClockClass, gs.ClockQuality.ClockAccuracy, ptpCfgName, BC)
```
nocturnalastro: It shouldn't need to do this because we get it from the `e.clkSyncState`

nocturnalastro: That's why there's a split between the two other clocks e.g T-GM don't have the state inside something like e.clkSyncState so they need to send it.

I think this will lead to two values being sent and we can't guarantee they are the same

pkg/event/event.go
jzding: same results. Gemini review recommends the latter one:
https://go.dev/play/p/cMD97nadfAn

nocturnalastro:
Your commit says that `EmitClockClass checks clkSyncState and finds it empty for ptp4l configs` I don't understand what this means. It just emits what's there. It shouldn't be empty when T-BC is running as that's the data for the state machine (https://github.com/k8snetworkplumbingwg/linuxptp-daemon/blob/main/pkg/event/event_tbc.go#L102). What's causing this to empty?

pkg/event/event.go
nocturnalastro: Ah ok because splitN is only splitting on the first one. I always find the N+1 in split N confusing.

pkg/event/event_tbc.go
vitus133: I think this should remain the lowercase announceClockClass. The uppercase AnnounceClockClass will send  the clock class taken from GRANDMASTER_SETTINGS_NP, which is sometimes wrong in the T-BC case

pkg/event/event.go
```diff
@@ -783,14 +784,19 @@ func (e *EventHandler) ProcessEvents() {
...
+						}
+						// Normalize to "ptp4l." so clock class is emitted under the ptp4l
+						// config name that cloud-event-proxy expects.
+						// Example: "ts2phc.1.config" -> "ptp4l.1.config"
+						key := strings.Replace(k, "ts2phc", "ptp4l", 1)
+						clkSnapshot[key] = v.clockClass
 					}
 					e.Unlock()
 					for clkCfgName, clockClass := range clkSnapshot {
-						parts := strings.SplitN(clkCfgName, ".", 2)
```
nocturnalastro, line 790: This being removed makes me nervous

pkg/event/event_tbc.go
nocturnalastro: I don't know if it will for T-BC I think it should hit the else in this if
https://github.com/openshift/linuxptp-daemon/pull/585/changes#diff-f35b8976e8af7ef3dd90969c4b2b48f8cdb23e77e2a79a2be4fea7225938e13dR771

pkg/event/event.go
jzding: I will put it back and do more testing today before I remove [WIP]. I have done so many iterations with git amend that I lost the original changes that actually tested working in the QE lab. I will ping you once I'm ready to review again.

jzding:
@verified later Bonnie-Block

PR #589, opened 2026-04-23
[release-4.22] OCPBUGS-84240: Fix T-BC clock class not re-emitted after cloud-event-proxy restart

jzding:
This is a cherry-pick of https://github.com/k8snetworkplumbingwg/linuxptp-daemon/pull/194

In T-BC mode, announceClockClass never called storeClockClassLocked, so clkSyncState was empty for ptp4l configs. After cloud-event-proxy restarts, EmitClockClass and the 60-second classTicker found nothing to re-emit, causing clock class metrics to be permanently lost.

Fix by adding storeClockClassLocked to announceClockClass so clkSyncState is populated on every clock class announcement. The classTicker's existing SplitN logic normalizes ts2phc config names to ptp4l for emission, so no changes are needed in event_tbc.go.


sebsoto:
/lgtm

@jzding any reason to not use the /cherry-pick command?

nocturnalastro (approved):
/lgtm

PR #595, opened 2026-04-28
OCPBUGS-83706: [4.22] Add E825 T-BC input pin configuration

vitus133:
Manual cherrypick to release-4.22
Enable REF0P and REF0N DPLL input pins when configured as T-BC.


nocturnalastro (approved):
/lgtm

PR #597, opened 2026-04-30
OCPBUGS-85014: OCPBUGS-79676: Fix hardware and ptp profile association

nocturnalastro:
Align hardware config - ptp config association logic with the new naming scheme, where profile names are prefixed with resource name sepaated by "_".


jzding (approved):
/lgtm

PR #608, opened 2026-06-09
OCPBUGS-87869: Allow phase offsets from Active state pins

vitus133:
Newer DPLL drivers differentiate between an input pin's administrative and operational states. When the operational state is supported, the legacy state maps to the administrative state. Because the administrative state is capped at "selectable", it fails the legacy pin state filter.

Add pin-operstate, measured-frequency, frequency-monitor to dpll-netlink Accommodate dpll kernel spec additions: new pin-operstate enum (active/standby/no-signal/qual-failed), measured-frequency pin attr in millihertz, frequency-monitor device feature flag, and operstate/ffo/ffo-ppt fields inside pin-parent-device nest. Extends ParseDeviceReplies, ParsePinReplies, all HR structs, and adds table-driven unit tests for every new decode path. Modify the filter to accept phase offsets from either:
- "connected" input pins (legacy drivers)
- Input pins with an "active" operational state (new drivers)

Assisted by Cursor

### DIFF
--- a/pkg/dpll-netlink/dpll-uapi.go
+++ b/pkg/dpll-netlink/dpll-uapi.go
@@ -265,22 +265,26 @@ func GetClockQualityLevels(cqs []uint32) string {
 
 // GetClockQualityLevel returns clock quality level as a string
 func GetClockQualityLevel(cq uint32) string {
-	clockQualityLevelMap := map[uint32]string{
-		ClockQualityLevel:             "itu-opt1-prc",
-		ClockQualityLevelITUOpt1PRC:   "itu-opt1-prc",
-		ClockQualityLevelITUOpt1SSUA:  "itu-opt1-ssua",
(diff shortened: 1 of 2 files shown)


pkg/dpll-netlink/dpll-uapi.go
```diff
@@ -135,6 +254,35 @@ const (
...
+	clockQualityLevels := make([]string, 0)
+	for _, cq := range cqs {
+		clockQualityLevels = append(clockQualityLevels, GetClockQualityLevel(cq))
+	}
+	return strings.Join(clockQualityLevels, ",")
+}
+
+// GetClockQualityLevel returns clock quality level as a string
+func GetClockQualityLevel(cq uint32) string {
+	clockQualityLevelMap := map[uint32]string{
```
nocturnalastro: I feel like this should be a switch

vitus133: Done

PR #609, opened 2026-06-09
OCPBUGS-86449: Add HPE GNR-D adaptations

vitus133:
Cherry-pick 4 commits for HPE GNR-D adaptations from main:

- Allow linking interfaces by PHC
  
    Previously, interfaces were logically linked by a common name prefix. For
    example, 'eno4' and 'eno5' were aliased as 'enox' and linked to the same
    PTP hardware clock (PHC) island. This link is required in Telecom Boundary
    Clock (T-BC) implementations to relate the upstream port to the leading
    NIC port that exposes the PTP device.
    
    However, some hardware implementations (such as the HPE ProLiant EL140 Gen12)
    use different prefixes for ports belonging to the same NIC,
    causing name-prefix association to fail.
    
    Fix this by allowing the leading NIC's default interface and the upstream
    port to be associated directly by their PHC.

-  Add HPE EL140 vendor defaults

- Fix interface alias store initialization
    
    Clear the store at the start of applyNodePTPProfiles() (after processes stop, so no events can race), and also uncomment the reset line inside  calculateAliases() so it's correct even if called from other paths

- Sanitize GNSS UART before starting gpsd
    
    On platforms where the GNSS module is connected via UART,
    killing gpsd during configuration transitions can leave
    the interface in an inconsistent state. This prevents gpsd
    from reconnecting properly on the next start.
    
    This commit adds an stty sane invocation prior to starting gpsd.
    This ensures the UART is in a clean operational state.
    No regressions were observed on USB-connected GNSS modules.

### DIFF
--- a/pkg/alias/alias.go
+++ b/pkg/alias/alias.go
@@ -32,6 +32,13 @@ func GetAllAliases() map[string]string {
 	return storeInstance.getAllAliases()
 }
 
+// GetPhcGroup returns the PHC device path for the given interface, or empty
+// string if the interface was not registered. Two interfaces with the same
+// non-empty return value share the same PHC hardware.
+func GetPhcGroup(ifname string) string {
+	return storeInstance.getPhcGroup(ifname)
+}
+
 // Debug ...
 func Debug(logF func(string, ...any)) {
 	for ifName, alias := range storeInstance.aliases {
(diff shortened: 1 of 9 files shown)


nocturnalastro (approved):
/lgtm

PR #616, opened 2026-06-23
OCPBUGS-82519: Fix chronyd NTP failover

vitus133:
stop/start chronyd process instead of chronyc offline/online

The chronyc offline command was unreliable — chronyd could re-acquire NTP sources despite being told to go offline, causing clock frequency interference with phc2sys (clockcheck events). Replace with process stop/start, using the same pattern as phc2sys.

The disable path uses `go p.cmdStop()` (non-blocking) to avoid a deadlock: cmdStop blocks on exitCh, but the ntpfailover FSM callback runs inside the process scanner goroutine which must finish before cmdRun can send to exitCh.


nocturnalastro (approved):
/lgtm

PR #623, opened 2026-07-08
OCPBUGS-97962: Fix worst-offset calculation

vitus133:
On WPC multi-NIC T-BC systems, follower DPLLs that are out-of-range set phaseOffset to FaultyPhaseOffset (99999999999) before sending events. This sentinel value was inserted into the shared DPLL window, corrupting window.Mean() and causing getLargestOffset to miss follower DPLL offsets — declaring Locked prematurely based only on ptp4l offsets.

Guard window.Insert() in Data.AddEvent to skip when offset equals FaultyPhaseOffset. The sentinel is still stored in dd.Offset so it participates in the worst-case abs() comparison in getLargestOffset, correctly blocking the Locked transition until followers converge.

### DIFF
--- a/pkg/event/stats.go
+++ b/pkg/event/stats.go
@@ -93,7 +93,9 @@ func (d *Data) AddEvent(event EventChannel) {
 				if fnd {
 					offset := off.(int64)
 					dd.Offset = offset
-					d.window.Insert(float64(offset))
+					if offset != FaultyPhaseOffset {
+						d.window.Insert(float64(offset))
+					}
 				}
 
 			} else {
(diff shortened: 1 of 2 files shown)


pkg/event/stats.go
```diff
@@ -89,12 +89,14 @@ func (d *Data) AddEvent(event EventChannel) {
...
 				dd.logData = event.GetLogData()
-				off, fnd := event.Values[OFFSET]
-				if fnd {
-					offset := off.(int64)
-					dd.Offset = offset
+			off, fnd := event.Values[OFFSET]
+			if fnd {
+				offset := off.(int64)
+				dd.Offset = offset
+				if offset != FaultyPhaseOffset {
```
lack: Is the indenting weird?

vitus133: fixed

PR #624, opened 2026-07-09
OCPBUGS-97827: Add ptp offset condition to enable DPLL

vitus133:
Backport of ptp4l offset condition for DPLL input enablement

### DIFF
--- a/pkg/daemon/daemon.go
+++ b/pkg/daemon/daemon.go
@@ -41,23 +41,25 @@ import (
 )
 
 const (
-	PtpNamespace                    = "openshift-ptp"
-	PTP4L_CONF_FILE_PATH            = "/etc/ptp4l.conf"
-	PTP4L_CONF_DIR                  = "/ptp4l-conf"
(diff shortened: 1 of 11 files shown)


pkg/daemon/daemon.go
```diff
@@ -216,9 +218,16 @@ func (p *ProcessManager) EmitClockClassLogs() {
...
+	trPortsConfigFile string
 	// Time receiver interface name for T-BC clock monitoring
-	trIfaceName string
+	trIfaceName        string
+	lastReportedState  event.PTPState
+	lastAppliedState   event.PTPState
+	offsetFilter       *utils.Window
+	offsetThreshold    float64
+	offsetEventWindow  *utils.Window
+	lastOffsetEventSec int64
```
vitus133, line 230: lastOffsetEventSec is not the last offset value. It's the Unix timestamp (in seconds) of the last time sendPtp4lOffsetEvent emitted an event to the channel. It rate-limits emissions to once per second:

nocturnalastro (approved):
/lgtm

PR #631, opened 2026-08-05
OCPBUGS-104578: Delay T-BC ts2phc startup until offset qualified

vitus133:
Manual cherry-pick
Hold ts2phc behind skipInitialStartup on T-BC until the offset-filter gate that enables DPLL inputs fires, and only after phc2sys is released. Leave T-GM ts2phc-first ordering and the phc2sys 1s delay unchanged.


nocturnalastro (approved):
/lgtm

PR #634, opened 2026-08-18
OCPBUGS-111603: force freerun when non-leading DPLL loses lock

vitus133:
In chained-NIC T-GM/T-BC topologies, a non-leading DPLL losing lock must degrade the composite clock to FREERUN (S0). Previously the clock incorrectly reported HOLDOVER (S1) because:
1. UpdateState() gave HOLDOVER priority over FREERUN
2. The event package had no leading vs non-leading DPLL awareness.
3. stateDecision() used source-specific branches (GNSS/PPS/PTP4l) that don't reflect actual topology in multi-NIC configs. Changes:
- Fix UpdateState() priority (stats.go)
- Add hasNonLeadingDPLLFault() to detect follower faults when the leading DPLL is still locked (event.go)
- Apply to both T-GM (updateGMState) and T-BC (updateBCState) paths
- Refactor stateDecision() HOLDOVER branch: unify GNSS/PTP4l under hasLeadingSource(), non-leading DPLLs go directly to FREERUN
- Add leader-follower matrix state tests (stats_test.go)

### DIFF
--- a/pkg/dpll/dpll.go
+++ b/pkg/dpll/dpll.go
@@ -763,45 +763,29 @@ func (d *DpllConfig) stateDecision() {
 
 	case DPLL_HOLDOVER:
 		switch {
-		case d.hasPPSAsSource():
+		case !d.hasLeadingSource():
 			d.state = event.PTP_FREERUN
 			d.phaseOffset = FaultyPhaseOffset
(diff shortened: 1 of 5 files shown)


nocturnalastro (approved):
/lgtm

PR #636, opened 2026-08-24
[release-4.22] OCPBUGS-109638: Fix clock_state metric: HOLDOVER mapped to 0, stale when CEP enabled

jzding:
## Summary
Backport of upstream PR https://github.com/k8snetworkplumbingwg/linuxptp-daemon/pull/250 to release-4.22.

Two fixes for the `openshift_ptp_clock_state` Prometheus metric:

- **HOLDOVER mapped to 0 instead of 2** — `updateClockStateMetrics` only distinguished LOCKED (1) from everything else (0). HOLDOVER was indistinguishable from FREERUN despite the help text declaring `0 = FREERUN, 1 = LOCKED, 2 = HOLDOVER`. Replaced the if/else with a switch using typed constants shared between `metrics.go` and `event.go`.

- **Metric stale when CEP enabled** — `EventHandler.UpdateClockStateMetrics` was guarded by `if e.stdoutToSocket { return }`, skipping Prometheus updates when cloud-event-proxy was enabled. Since CEP does not register its own `openshift_ptp_clock_state` gauge, this left the metric stale. Removed the guard so the metric is always updated.

## Test plan
- [x] `TestUpdateClockStateMetrics` — verifies FREERUN→0, LOCKED→1, HOLDOVER→2
- [x] `go build ./...` passes
- [x] Lab tested on cnfdg32 (OCP 4.22) with upstream PR

### DIFF
--- a/pkg/daemon/daemon_internal_test.go
+++ b/pkg/daemon/daemon_internal_test.go
@@ -25,6 +25,8 @@ import (
 	"github.com/k8snetworkplumbingwg/linuxptp-daemon/pkg/leap"
 	ptpv1 "github.com/k8snetworkplumbingwg/ptp-operator/api/v1"
 	ptpv2alpha1 "github.com/k8snetworkplumbingwg/ptp-operator/api/v2alpha1"
+	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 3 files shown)


vitus133 (approved):
/lgtm

